Issue #75, opened 2020-03-21
bug: Event without signup form requires signup times


andreedd:
**Describe the bug**  
When turning off or setting an event without signup page to "Published" you have to fill out the event sign up times etc.. 
**To Reproduce**  
Create an event without signup forms and turn "published off"
**Screenshots**  
![Screen Shot 2020-03-21 at 14 54 21](https://user-images.githubusercontent.com/33100278/77227736-1e3a8800-6b8b-11ea-9983-c1778bcd07f2.png)
**Desktop, phone (please complete the following information):**  
 - OS: [e.g. iOS, Windows, Linux]

**Additional context**

oj-lappi:
I can take this one.
I think the behaviour we want is: if the sign up option is set to false, no checks against sign up dates.
(Or at least no way for the sign up dates to invalidate an edit).

oj-lappi:
The root cause seems simple enough: 
If we turn sign up off, then all the sign up fields are set to falsy values:

`events/forms.py`:

```       
if not post.sign_up:
            post.sign_up_max_participants = 0
            post.sign_up_others = None
            post.sign_up_members = None
            post.sign_up_deadline = None
            post.sign_up_cancelling = False
            post.sign_up_cancelling_deadline = None
```

Is there a reason for doing this?

oj-lappi:
I pushed a fix which makes the fields not required (in the edit form).

Will there be a problem if the sign up fields are empty?

Faagerholm:
Should be tested. Does it affect any fields once a user is registering for an event? What happens if we update an event by including the signup?

Faagerholm:
This has been tested and works as intended, :+1:

PR #80, opened 2020-03-22
Mby fix issue #75: makes signup fields nonrequired

oj-lappi:
Fixes the immediate issue, but we may want to think about this a bit more.
Is it important that the signup fields have values?


oj-lappi:
Fixes the immediate bug for issue #75 
Buuut, is there problems if the signup fields are empty?

PR #107, opened 2020-07-11
Date/kkmerge

andreedd:
Big merge from with many features from KK/develop 

Restyled from KK to DaTe

Mby needs some review?


nDny (approved):
Works on my machine after some minor edits are made to accomodate the new changes.
This will have to be done to the production site as well.

Just a little mention about imports, right now they are not very consistent. As the project grows it can get messy so easier to sort it sooner than later.
Imports in django should be grouped like [this](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/).

image
Example in date/views.py

Faagerholm:
Tested both from a new installation and on top of an old one. As long as all migrations are done correctly the project should start up.

The Static nav group could be reorganised to just show the static pages nav -> under this the list of all groups should be shown and new items would be added under a given group.

A very quick code review was done.

Faagerholm (approved):

Faagerholm:
I think we can be happy with 2 PR reviews. If both @nDny and @andreedd are happy with the changes.

PR #112, opened 2020-07-24
kkmerge to QA

andreedd:


Faagerholm (approved):
As this has already been tested in-dept earlier we can safely merge this.

Check #107 for more information.

Issue #138, opened 2020-11-25
feat: blog module


Faagerholm:
**This is a test issue for our new officials**

The issue is created to get familiar with Github and our workflow in this project.

---
As a new official
- Assign this issue to yourself.
- Write a hello comment to this issue.
- Check our this [wikipedia page](https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/wiki/Meeting-26-11-20) for more information
- Tag this issue when committing to your new feature branch and when you create your final PR.

WilliamNylund:
a hello comment

atokos:
hello, a hello comment

rkantero:
a hello comment to this issue.

vsjostro:
issue to this comment hello

hmartola:
hello

jaspnas:
No longer used, closing as completed.

PR #141, opened 2020-11-28
Feature/blog module william

WilliamNylund:
You can also tag an issue in your PR comment. 
Like this: #138 <- The will appear on the issue. 

De e najs!


blog_module/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+app_name = 'blog_module'
+
+urlpatterns = [
+    path('', views.index, name='index'),
+    path('blogs/<slug:blog_id>/', views.detail, name='detail'),
```
Faagerholm, line 9: This will render out to `.../blog_module/blogs/<id>`

the URL could be shortened, but that's an opinion on taste.

blog_module/views.py
```diff
@@ -0,0 +1,12 @@
+from django.shortcuts import render
+
+from . import models
+
+def index(request):
+    all_blogs = models.Blog.objects.all()
+    return render(request, 'index.html', {'all_blogs': all_blogs})
```
Faagerholm, line 7: Another way to write this is to create a context dict. It's easier to extend and read

```python
...
context = {
   'all_blogs': all_blogs,
   # later more context can  be added here
}
 return render(request, 'index.html', context)
```

PR #168, opened 2021-04-27
Feature/s3 storage

andreedd:
Working archive with minio.
Upload script for old images.
Pagination for image gallery.


archive/mass_upload.py
```diff
@@ -0,0 +1,81 @@
...
+logger = logging.getLogger('date')
+
+logger.info("STARTING IMAGE UPLOADER")
+
+def s3_config():
+    return boto3.client('s3',
+            endpoint_url=settings.AWS_S3_ENDPOINT_URL,
+            aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+            region_name='eu-north-1')
```
Faagerholm, line 30: Could this be static? Could be done later

Faagerholm (approved):
Did not have time to test this. But looks good to me :)

archive/mass_upload.py
```diff
@@ -0,0 +1,81 @@
...
+cursor = connection.cursor()
+
+query = ''' INSERT INTO archive_picture 
+        (image, favorite, collection_id) 
+        VALUES (%s,%s,%s) '''
+
+cursor.executemany(query, query_list)
+transaction.commit()
+
+logger.info("IMAGE UPLOADER COMPLETE")
```
andreedd, line 81: This is not the most optimal way to do it, but for now it will do.
We figured that the most optimal way would be to have 2 separate scripts.
One which is run locally where the images are located, 
and one which is run on the server to insert collections and images directly into the database by checking the images from minio via the AWS SDK (boto3). The images in minio should have the path /media/<year>/album_name>/<file_name>

PR #169, opened 2021-04-28
s3 storage with minio

andreedd:
S3 storage with minio to testing on QA


andreedd:
And some other small changes

PR #210, opened 2022-01-20
Add validation for unique emails on event registration form

### LINKED ISSUE #189: fix: Signup with same email is not added to list event attendees
**Describe the bug**  

People can at the moment send the event form without being added to the event attendee list if they sign up with a email that is already used  by another event attendee, without being notified that their signup didn't go through. 

**To Reproduce**  

Sign up to event, then sign up again with same email

**Additional context**  

We should have some kind of validation when pressing the "anmäl" button when trying to sign up with an email which has already been signed up with

WilliamNylund:


events/static/events/css/form.css
```diff
@@ -18,6 +18,10 @@
     border-radius: 5px;
 }
 
+.form > ul.errorlist{
+    display: none;
+}
+
```
WilliamNylund, line 24: Hiding the built in djangos BaseForm error messaging since we've added our own

image

andreedd (approved):
Seems to be working well, good commit!

PR #217, opened 2022-02-10
Feat/website initialization

### LINKED ISSUE #211: feat: Add base fixtures for development
We should have some predefined data for development so we don't have to add manually each time

nDny:
Adds a little shell script that enables a user to wipe and restore their migrations and database to a "stock" qa test state.

Fix #211


Faagerholm:
Could this be also done by simply removing the docker volume, and then running the initialdata.json on the newly created volume?

Django has a built-in tool to reverse migrations as well. 
`python manage.py migrate <App-name> zero` will remove all migrations for that app. Maybe something that could be used? 
https://docs.djangoproject.com/en/4.0/topics/migrations/#reversing-migrations

scripts/clean_init.sh
```diff
@@ -0,0 +1,47 @@
...
+
+echo "WARNING! This will restore your project to an intial state"
+read -p "Are you sure? y/n " -n 1 -r
+echo    # (optional) move to a new line
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1 # handle exits from shell or function but don't exit interactive shell
+fi
+
+source ../example.env
```
Faagerholm, line 12: Do we want to source the example.env file? will this not overwrite our $USER.env variables?

Faagerholm (approved):
Looks good 👍, a question about using the example.env as a source, but cannot say if that is intended so will approve.

scripts/clean_init.sh
nDny: Had it in there in case the user did not source the env file before running. 

This is a very rough script though so can definitely change it if needed (do a find check to see if $USER.env exists for example), but wrote it with the mindset that if you run this you either way reset everything.

Could probably add a help text to the script or in the readme in case you have a custom env that you dont want overridden.

nDny:
> Could this be also done by simply removing the docker volume, and then running the initialdata.json on the newly created volume?
> 
> Django has a built-in tool to reverse migrations as well. `python manage.py migrate <App-name> zero` will remove all migrations for that app. Maybe something that could be used? https://docs.djangoproject.com/en/4.0/topics/migrations/#reversing-migrations

The volume deletion thing is definitely a viable option, only difference in doing it this way is that nothing has to be recreated if you already have one (which essentially has the same end result anyways). 

And the django migration reset would make sense if it were only one app we were resetting migrations for. Afaik you would have to do it app-by-app with the `migrate zero` command. This way we don't have to target anything specifically and can just nuke everything :D.

Faagerholm:
Here's a quick one-liner I found on SOF, tho the same comment says it's easier to nuke the DB 😄 

`python manage.py showmigrations | grep -P "^[^ ]" | xargs -I {app} python manage.py migrate {app} zero`

[source](https://stackoverflow.com/a/63275306/10389764)

PR #219, opened 2022-02-12
add pagination

### LINKED ISSUE #202: fix: Paginate news
Currently only n amount of news are shown at /news

We should paginate the news posts at /news so that we can browser all previous news and not have to show all previous news on same page

**Additional context**  

https://docs.djangoproject.com/en/4.0/topics/pagination/ could use this probably

hmartola:
the pagination is similar to the one found in archive app


andreedd (approved):
Looking good.
Added latest css from master and also added same code to Albins angels news in view.

PR #220, opened 2022-02-13
Feature/avec

### LINKED ISSUE #199: feat: "Avec" functionality for events
We should have a way to sign up two participants for an events at once.

**Additional context**
Some hacky way was done in KK Baal code.

WilliamNylund:
Added a second commit with functionality for choosing to hide registration preferences for avec form (if we want that). That would allow us to hide fields like bordssällskap from avec form.

### DIFF
--- a/events/admin.py
+++ b/events/admin.py
@@ -53,6 +53,12 @@ class EventAttendeesFormInline(OrderableAdmin, admin.TabularInline):
         readonly_fields = ['time_registered']
         return readonly_fields
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        event_id = request.resolver_match.kwargs.get('object_id')
+        if db_field.name == "avec_for":
+            kwargs["queryset"] = EventAttendees.objects.filter(event=event_id)
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
+
 @admin.register(Event)
 class EventAdmin(admin.ModelAdmin):
     save_on_top = True
--- a/events/admin.py
+++ b/events/admin.py
@@ -21,7 +21,7 @@ class EventRegistrationFormInline(admin.TabularInline):
     fk_name = 'event'
     extra = 0
     readonly_fields = ('line_number',)
-    fields = ('line_number', 'name', 'type', 'required', 'public_info', 'choice_list')
+    fields = ('line_number', 'name', 'type', 'required', 'public_info', 'hide_for_avec', 'choice_list')
     can_delete = True
 
     def line_number(self, obj):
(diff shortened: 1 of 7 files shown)


events/admin.py
```diff
@@ -38,14 +38,22 @@ class EventAttendeesFormInline(OrderableAdmin, admin.TabularInline):
...
     ordering = ['attendee_nr']
 
+    def get_fields(self, request, event):
+        fields = ['attendee_nr', 'user', 'email', 'anonymous', 'preferences', 'time_registered']
+        if event and event.sign_up_avec:
+            fields.append('avec_for')
+        return fields
+    def get_readonly_fields(self, request, event):
+        readonly_fields = ['time_registered']
+        if event and event.sign_up_avec:
```
nDny: The `avec_for` field is returned both as a normal field and a readonly field, which results in it being readonly in the admin pages. Unsure if you intended for this to be editable or not but should probably return it as either or. (Or if it actually needs to be in both places and I just missed why let me know :D)

I think it makes sense to have it editable in case you want to manually assign avec things from the admin page.

events/models.py
```diff
@@ -168,6 +188,7 @@ class EventAttendees(models.Model):
     preferences = JSONField(_('Svar'), default=list, blank=True)
     anonymous = models.BooleanField(_('Anonymt'), default=False)
     time_registered = models.DateTimeField(_('Registrerad'))
+    avec_for = models.ForeignKey("self", verbose_name=_('Avec till'), null=True, on_delete=models.SET_NULL)
```
nDny: To avoid potentially weird errors about this field being required in forms, it would make sense to also add a `blank=True` to this field.

nDny (commented):
Seems to work well :fire: 

Minor comment about the fields for the admin page but it is quite minor :sunglasses:

nDny:
Probably needs some tuning if we want it to work with @andreedd's recent årsfest changes, but you can probably look at that together.

events/admin.py
WilliamNylund: Thought that would make more sense as well, But apparently it's required since readonly fields has to be included in normal fields to be displayed.
https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields

I made it readonly because I didn't really see any reason to edit it, since people not using avec will use bordssällskap. But you're right, there will probably be cases where we want to edit it so I'll fix it

events/models.py
WilliamNylund: Nice, I'll add it.

WilliamNylund:
Should be working with @andreedd arsfest now.

andreedd (approved):
Clean implementation of avec.
Couldn't have done it better.

PR #233, opened 2022-02-21
Allow commodore to register peeps

nDny:


nDny:
Quick thing to allow årsfest committee to enter the reserved spots before signup

PR #254, opened 2022-11-10
Add basic handling for null events

### LINKED ISSUE #247: bug: events crash if date fields are empty
See images.

Members only was on. The crash happened for non-members.

Screenshot 2022-10-11 at 11 47 11
Screenshot 2022-10-11 at 11 51 29

nDny:
We should still add relevant texts in the "anmälning" part.

But for now this is a quick fix so we avoid crashes.

Fixes https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/issues/247

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -110,12 +110,18 @@ class Event(models.Model):
             registration.delete()
 
     def registration_is_open_members(self):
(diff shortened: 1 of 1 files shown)


andreedd:
Nice!

PR #264, opened 2023-01-19
Reversed upload order for images

JoarSabel:
Reversed the list of images to be uploaded so that they don't come out in the wrong order on the webpage. Seems to work at least when testing locally. 

This should not mess with the pre-existing albums that were uploaded in reverse order to compensate for the fault in question.


JoarSabel:
Changed it to reverse the image archive if the year is not 2022, since images from that year are uploaded in reverse in order to be correct. Also some stray migration in there that just hadnt been committed?

PR #269, opened 2023-01-19
Add a length limit to event text inputs

### LINKED ISSUE #268: fix: No input limit on events signup fields
Add a limit on input size for the events signup form. 

A server side check is critical but client side validation may be nice to have.

jaspnas:
Fixes #268 .


andreedd (approved):
Yeah nice this should fix it!

PR #272, opened 2023-01-20
Fixed issue: #267

### LINKED ISSUE #267: fix: Don't show event member limit if the isn't one
Website currently shows members limit for an event if there isn't a limit on the number of participants.
Change this to hide members limit if this is the case

image

Endorias:
changed "Ingen Begränsning" instead of 0 when there isn't a specified amount of participants 

Issue number #267


andreedd (commented):
Looks good, seems to be merge conflicts atm, so if you want to merge latest develop and fix those :D

Endorias:
I added an empty line by accident to members it seems (nothing to do with the fixed issue) and didn't notice it when committing. Should be fine now?

PR #283, opened 2023-01-26
Add Jet-dashboard to admin page

JoarSabel:
Implemented Jet-dashboard for the `/admin` page, this should probably be tested a bit more thoroughly. Maybe we use Styssen as test-animals?
It is possible to implement Google analytics, but do we even need it?


andreedd:
Yeah would probably be good to have a few people who are actively using the admin to test it. It really comes down a lot to personal preference.

PR #286, opened 2023-02-03
Update dependencies

### LINKED ISSUE #261: fix: Update dependencies
Dependencies are out of date. 

Python, Django and Postgres should be updated, probably Redis too. 

The code probably need to be changed quite a bit because of deprecation. There should also be a script to upgrade the Postgres database. 

Also a good idea to update the documentation to make it easier for future contributors to manage and update dependencies.

jaspnas:
# WIP PR for updating dependencies

Update for dependencies of the project

Closes #261 

### Things to do:

- [x] Update Python to 3.11
- [x] Move to alpine to reduce size of web container
- [x] Update Postgres to v15.x
- [x] Make a script to update the database automatically
- [x] Update Redis to version 7 and switch to the alpine image. 
- [x] Update Django to version 4.x
- [x] Update code to replace stuff deprecated in Django 4.0
- [x] Bump other python packages
- [x] Fix websocket or re-add asgiref lock

### The following things will need testing:

- [x] The database update script (Critical as this can lead to data loss)
- [x] Stability
- [x] Functionality of code that has been updated


jaspnas:
The database update script has not yet had much testing. It should at this point NOT be considered 100% stable.

jaspnas:
The update broke some language middleware stuff using the deprecated translation.LANGUAGE_SESSION_KEY. The middleware does not currently seem to have any effect on the website but probably will if we decide to enable translation. 

Does this need to be fixed in this PR or should I create a separate issue?

Django also throws a bunch of warnings: https://gist.github.com/jaspnas/f4220fc96493ab2de398a514a6bbd7f1

requirements.txt
```diff
@@ -1,5 +1,5 @@
-Django==3.1.13
-asgiref==3.2.10
```
andreedd, line 2: I have a small memory that something about the live attendance list at events broke at a certain version of asgiref & channels. We will probably have to test the websocket again and/or make changes to it. Or remove it or something :D

andreedd:
> The database update script has not yet had much testing. It should at this point NOT be considered 100% stable.

Probably best to also run the script at a populated DB at our QA server before. 👍

andreedd:
> The update broke some language middleware stuff using the deprecated translation.LANGUAGE_SESSION_KEY. The middleware does not currently seem to have any effect on the website but probably will if we decide to enable translation.
> 
> Does this need to be fixed in this PR or should I create a separate issue?
> 
> Django also throws a bunch of warnings: https://gist.github.com/jaspnas/f4220fc96493ab2de398a514a6bbd7f1

Let's fix this in another PR and issue. Probably best to fix this while actually implementing the translations

jaspnas:
> > The database update script has not yet had much testing. It should at this point NOT be considered 100% stable.
> 
> Probably best to also run the script at a populated DB at our QA server before. +1

I will still try to add some checks to the script that should lessen the risk of data loss if the container fails. It is not setup to delete the generated database dump so the critical part is the generation of the database dump.

requirements.txt
jaspnas: Websocket does indeed not seem to be working. I will look into that, added to PR tasks.

jaspnas: Websockets should now work, I added daphne as an additional requirement.

jaspnas:
I have added a check to the script that should prevent data loss from the container failing to start. 

PR should be mostly feature complete, marking ready for review even though it still needs testing

nDny:
We will most likely need to do add this to our settings with the Django 4.0 changes:
https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS

andreedd:
> We will most likely need to do add this to our settings with the Django 4.0 changes: https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS

Yes a CSRF_TRUSTED_ORIGINS = ORIGINS_LIST_VARIABLE was needed

update-postgres.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+# Check if the required environment variables are set
+if [ -z "$DATE_POSTGRESQL_VERSION" ] || [ -z "$DATE_DB_PORT" ] || [ -z "$DATE_DB_PASSWORD" ]; then
+  echo "Error: Required environment variables are not set"
+  exit 1
+fi
+
+# Make sure website is stopped
+docker compose down && sleep 15 && docker-compose up db -d
```
andreedd: Our server still runs an old version of Docker and "docker compose" will need to be switched to "docker-compose"

jaspnas: Should be fixed in latest commit

example.env
```diff
@@ -33,6 +36,9 @@ export DEFAULT_FROM_EMAIL='admin@datateknologerna.org'
 # Allowed hosts
 export ALLOWED_HOSTS='["0.0.0.0", "127.0.0.1", "localhost"]'
 
+# Allowed origins for csrf
+export ALLOWED_ORIGINS='["https://localhost", "http://localhost"]'
```
jaspnas, line 40: Should this be set to something else by default?

andreedd: I think it can actually be set to an empty list during development, e.g. []

andreedd: Might be wrong but

jaspnas: IIRC it should work with the empty list in development.

I was thinking of having the default values work an example as I don't think they have any effect in development.

andreedd (commented):
Yes

andreedd (approved):
Should work

PR #290, opened 2023-02-08
Add admin password reset and reformat members module

### LINKED ISSUE #193: feat: password reset in admin
**Is your feature request related to a problem? Please describe.**

Seems like sometimes people have problems with receiving the email. The person did not receive the email while when we tried it was received like it should have.

A field in the admin "Reset Password"  where you could write a new password for the user.

**Additional context**

Only board members and superadmin should have access to users in admin so should be ok.

jaspnas:
Added ability to change user password to members admin. 
Fixes #193 

Also reformats the code in members and removes unused imports. 

Password reset is currently implemented as a read only info field and a link to a separate default password change form.

image
image


andreedd:
Fix conflicts in this and we can move this one and #302 to QA 😃

PR #298, opened 2023-02-16
Fix environment variables for biocum

jaspnas:
Sets envs for biocum correctly and changes date- aliases in example.env to cum-


jaspnas:
Part of #316

PR #300, opened 2023-02-16
Add memeber-cert-page for supporting members

### LINKED ISSUE #282: feat: Stödjande medlemmar should always have a red members pass
Screenshot 2023-01-25 at 10 54 52

Request from stysse.

Supporting members should be able to sign up for members only events but their members pass should be red.

JoarSabel:
Made the member cert page visible to supporting members, and made the color red for supporting members, although do we want it red or should we have a separate color?

Closes #282


andreedd:
Made other issue for color: #303 
Merging this

PR #312, opened 2023-03-22
Fixed hardcoded checks in archive index

### LINKED ISSUE #23: fix: Replace the hard coded group name in archive module
Archive requires a user to be in group "fotograf" so they can upload galleries and images.

JoarSabel:
Not sure if this is what was meant with https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/issues/23 but I have consumed a lot of RUM and did something, idk.

Closes #23

### DIFF
--- a/archive/templates/archive/index.html
+++ b/archive/templates/archive/index.html
@@ -13,7 +13,7 @@
         <div class="container-size" style="max-width: 600px;">
             {% if type == 'pictures' %}
                 <h1 class="text-center">&nbsp; {% trans "Bildarkiv" %} &nbsp;</h1>
-                {% if request.user|in_group:"fotograf" or request.user|in_group:"styrelse" %}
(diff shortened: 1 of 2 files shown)


jaspnas:
Should this instead check if the user has the correct permissions for uploading pictures (for expample `archive | bild | Can add bild`) instead of them being in a group named "fotograf"?

nDny:
> Should this instead check if the user has the correct permissions for uploading pictures (for expample `archive | bild | Can add bild`) instead of them being in a group named "fotograf"?

This is not wrong but feels kind of redundant since no one should ever upload pictures except for people in fotograf or styrelse, the way it is in this PR makes that restriction clearer.
Or if we for some reason want to further target individual permissions which would have overlaps with these groups.

jaspnas (approved):
Looks good

PR #314, opened 2023-03-23
Serverless image resizing

jaspnas:
Implements usage of a digitalocean serverless function to reduce image file size

TODO:

- [ ] Switch the DigitalOcean URL to DaTes account
- [ ] Evaluate if the serverless function should save a version of the image back to digitalocean


KuRR221 (approved):

jaspnas:
Redo later

PR #315, opened 2023-03-27
KK version upgrade

jaspnas:
Version upgrade for KK

Includes changes from #286, #290, #299, #302, #306 and #310.


date/functions.py
```diff
@@ -11,7 +11,3 @@ def slugify_max(text, max_length=50):
         return trimmed_slug
     # First word is > max_length chars, so we have to break it
     return slug[:max_length]
-
-
-def days_hence(days):
-    return timezone.now() + timezone.timedelta(days=days)
```
JoarSabel, line 17: image
What the hell was this

archive/mass_upload.py
JoarSabel, line 1: Spoke with @jaspnas and it seems there are no references to mass_upload.py in the code, in which case, if this is a script. should we move it to scripts at some point?

JoarSabel (approved):
Looks good to me, should test on QA 

image

date/functions.py
jaspnas: Removed because pycharm showed it not having usages

archive/mass_upload.py
jaspnas: Moved to #319

PR #327, opened 2023-05-08
Feat/alumni signup

nDny:
Alumni request to have ARG members sign up on the DaTe page instead of ASG.

Basically an improved copy of the ASG signup stuff.


andreedd (approved):
Ye should be good

PR #328, opened 2023-05-08
Fix/field bug

nDny:
Fix breaking changes in django 4.2.1

Also add helper classes to not make people suicidal when only being able to upload 1 image at a time.


archive/forms.py
```diff
@@ -3,24 +3,40 @@
...
+            result = [single_file_clean(d, initial) for d in data]
+        else:
+            result = single_file_clean(data, initial)
+        return result
+
+
 class PictureUploadForm(forms.Form):
     album = forms.CharField()
-    images = forms.ImageField(required=False, widget=forms.ClearableFileInput(attrs={'multiple': True}))
+    images = forms.ImageField(required=False, widget=forms.ClearableFileInput())
```
andreedd: Should it use the MultipleFileInput?

nDny: Sure, but does anyone even use that one lol :D Is that not the one obfuscated somewhere in the archive view

nDny: Yup this one 
image

jaspnas: > Sure, but does anyone even use that one lol :D Is that not the one obfuscated somewhere in the archive view

I think I saw someone use that one at some point

archive/forms.py
```diff
@@ -3,24 +3,40 @@
...
 
 
 class ExamUploadForm(forms.Form):
     title = forms.CharField()
-    exam = forms.FileField(required=False, widget=forms.ClearableFileInput(attrs={'multiple': False}))
+    exam = forms.FileField(required=False, widget=forms.ClearableFileInput())
```
andreedd: Was thinking for these to use multiple as well? `MultipleFileInput`

archive/forms.py
```diff
@@ -36,7 +52,7 @@ def save(self, *args, **kwargs):
 
 
 class DocumentAdminForm(forms.ModelForm):
-    files = forms.FileField(widget=forms.ClearableFileInput(attrs={'multiple': True}),
+    files = forms.FileField(widget=forms.ClearableFileInput(),
```
andreedd: Was thinking for these to use multiple as well? `MultipleFileInput`

archive/forms.py
```diff
@@ -54,7 +70,7 @@ def save(self, *args, **kwargs):
         return collection
 
 class PublicAdminForm(forms.ModelForm):
-    files = forms.FileField(widget=forms.ClearableFileInput(attrs={'multiple': True}),
+    files = forms.FileField(widget=forms.ClearableFileInput(),
```
andreedd: Was thinking for these to use multiple as well? `MultipleFileInput`

andreedd: This is for uploading public s3 files, used when making news posts with public images for example

archive/forms.py
andreedd: This so Hektor does not have to upload single :D

nDny:
Yeh probably makes sense, I only put it on the most critical one because Django docs said that it **_should_** be a temporary fix and that they will implement a proper multi choice input later.

andreedd (approved):

PR #337, opened 2023-06-19
Feat/pip compile + arm compat

### LINKED ISSUE #336: feat: Handle dependencies with pip-compile
Would be nice to not have to manage dependencies manually.

Maybe switch to a tool like [pip-compile](https://github.com/jazzband/pip-tools) to handle them for us.

jaspnas:
Closes #336 

Allow easily locking dependencies with pip-compile

Also fixes ARM compatibility.


andreedd:
Sure, could try it out.

Maybe add to the readme some examples of using it?

jaspnas:
Would maybe be a good idea to have Github actions regenerate the dependencies once a month and create a pull request, Then the smoketest-workflow from #344 should be enough to see if it breaks anything

andreedd:
Yeah that could work, could even go as far as have actions do a commit with the dependencies at the same time if we want to.

jaspnas:
> Yeah that could work, could even go as far as have actions do a commit with the dependencies at the same time if we want to.

Do we want to manually review it? Are we fine with breaking QA if we set up a CD pipeline from develop if the dependencies break something?

andreedd:
I would say it is fine, since if we always commit them we can always revert them as easily as well.

jaspnas:
Conflicting, will redo later

PR #342, opened 2023-06-25
Remove dotenv

### LINKED ISSUE #341: Remove core/.env
To get a working production environment core/.env needs to be modified. This shouldn't happen as the file is checked into version control and may interfere with updates.

jaspnas:
Closes #341 

Removes `core/.env` and cleans up `example.env`


core/settings.py
```diff
@@ -266,7 +264,7 @@
 EMAIL_HOST = env('EMAIL_HOST')
 EMAIL_HOST_USER = env('EMAIL_HOST_USER')
 EMAIL_HOST_PASSWORD = env('EMAIL_HOST_PASSWORD')
-EMAIL_PORT = env('EMAIL_PORT')
+EMAIL_PORT = 587
```
andreedd, line 278: Any specific reason why not to include port nr as variable?

jaspnas: Port 587 should usually be the default for sending email with TLS. Most servers that we would use for sending mail will probably support 587 so I don't think it's worth adding extra options to the config that nobody will touch. It should be easy to add back later if needed.

andreedd: Yeah sure, probably won't have to use any of the other ports.

andreedd (approved):

jaspnas:
Once merged requires manual work to get this working. 

Holding until someone has time for it

jaspnas:
Replaced by #433

PR #343, opened 2023-06-25
Add unapplied migration

jaspnas:
Add a migration for alumniemailrecipient


nDny (approved):
Ah seems I forgot to make the migration with the final changes in the pr 😄 good catch

Issue #340, opened 2023-06-21
feat: CI jobs


jaspnas:
We should at minimum have something that automatically tests that the site starts up properly on each pull request

We will probably use GitHub actions for this as they are free/cheap and they don't require any maintenance on the infra side.

Later maybe set up some jobs to deploy them to qa/prod automatically once they get pushed to the correct branch. 
This should probably be integrated into #330

jaspnas:
A CI pipeline for testing exists and will be added to when needed

PR #344, opened 2023-06-25
Add simple workflow for checking website startup

jaspnas:
#340 

Supposed to be run only as an early warning for failures preventing startup


jaspnas:
Please don't merge before workflow reliably succeeds

andreedd (approved):
Should be good enough 👍

PR #351, opened 2023-07-29
Add workflow for building docker image

### LINKED ISSUE #347: feat: Build container images automatically and pull from registry
If we want to be able to scale across multiple nodes and improve reproducibility we should have a Github actions pipeline to build images on deploy/schedule to update dependencies that we pull when we want to start a container. 

Options for the container registry are for example:
- Github container registry
- ECR or other managed container registry
- self-hosting a container registry

jaspnas:
Adds image build and uploads it to gchr

Do we want to use gchr or some other registry?

Closes #347

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*.env
+*.md
+media/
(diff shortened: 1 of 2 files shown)


andreedd:
Looks ok to me.

I think It is okay to use GHCR.

Should be ok for the image to be public.

andreedd (approved):
ok

Issue #354, opened 2023-08-15
bug: Uploading pictures to homepage not working


KuRR221:
Cannot upload pictures via admin page or pictures page.

To reproduce try to upload picture from admin page in album and hit save.

![el mao](https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/assets/81564668/300740a3-9976-4187-8843-f6af1f1fc064)

Desktop, Windows 11, Chrome browser.

jaspnas:
Probably caused by #328

PR #356, opened 2023-08-15
KK: Fix pil deprecating Image.ANTIALIAS

jaspnas:
KK fix for #354


andreedd (approved):
OK if it seem to work

PR #358, opened 2023-08-17
Store alumni personal information in admin instead of sending email

### LINKED ISSUE #353: CRITICAL: Possible personal information sent through email
During signup to ARG emails are sent unencrypted containing personal information. 

This should be changed so that the email e.g. sends a link to an admin page instead and sends the minimum amount of data possible through mail. 

Issue originates from #327

jaspnas:
Fixes #353 

Need give ARG admin accounts before this is merged
~~Also need to check permissions for member module~~  works


jaspnas:
Currently seems like ARG admins would get access to image gallery/events etc.

Need to figure out if this is a problem

members/templates/alumni_signup_email_admin.html
```diff
@@ -1,11 +1,5 @@
...
-    Tel nr: {{ alumni.phone_number }}
-    Inskrivningsår: {{ alumni.year_of_admission }}
-    Arbetsplats: {{ alumni.employer }}
-    Arbetsuppgift: {{ alumni.work_title }}
-    Adress: {{ alumni.address }}
-    TFiF medlemskap: {{ alumni.tfif_membership }}
-    Får kontaktas: {{ alumni.alumni_newsletter_consent }}
+    {{ alumni.name }} har registrerat sig till ARG.
+
+    Full information: https://datateknologerna.org/admin/members/alumnisignup/{{ alumni_id }}
```
andreedd, line 4: This URL does not seem to exist in our URLS?
Or do we have a issue to create it?

jaspnas: You should get it to work in testing env by replacing datateknologerna.org with your dev server url (localhost:8000)

It's an admin page in the members module with the personal information of the alumni.

andreedd: Oh duh right it was admin msg and page 😄 Yeah should be good

andreedd (approved):
Overall I think this should be ok, one change we could make is to put the texts from the form in the models since we changed it to modelform. Then we don't have to define the fields again.

nDny:
~~Won't this make the signup information be available on the open web (simply behind the right URL) essentially just moving the GDPR issue to another place instead and making it accessible to everyone.~~

Edit: nvm, when I looked at it briefly I thought we sent the information to the person signing up lol (apparently me and DD looked at it with the same eyes) 😄 one could think I would remember what I made myself 🥲

nDny (approved):

jaspnas:
Needs testing on QA

PR #363, opened 2023-09-05
Kk/develop -> KK/Main

jaspnas:
Waiting for testing on QA


andreedd (approved):
Tested.

Image upload works.
Alumni signup seems to work. Email is sent and received and registration is added to admin.

Migrations should be ok.

PR #364, opened 2023-09-10
change fontawesome kit to datedatorer owned

andreedd:
Limited to:

* local development
* datateknologerna.org
* kemistklubben.org
* biologica.fi


jaspnas (approved):
Shouldn't break anything

PR #368, opened 2023-09-16
Store alumni personal information in admin instead of sending through…

jaspnas:
… mail (#358)

Access to ARG admins needs to be granted before merging


nDny (commented):
Now when using a model we can clean up the form class a lot.

```python
from django.utils.translation import gettext_lazy as _


class AlumniSignUpForm(ModelForm):
    class Meta:
        model = AlumniSignUp
        exclude = ["signup_date", "acknowledge"]
        labels = {
            "name": _("Namn"), 
            rest of fields...
        }
        help_texts = {
            more fields
        }
```

Not tested but something like that 😄

nDny:
Otherwise 👍

jaspnas:
> Now when using a model we can clean up the form class a lot.
> 
> ```python
> from django.utils.translation import gettext_lazy as _
> 
> 
> class AlumniSignUpForm(ModelForm):
>     class Meta:
>         model = AlumniSignUp
>         exclude = ["signup_date", "acknowledge"]
>         labels = {
>             "name": _("Namn"), 
>             rest of fields...
>         }
>         help_texts = {
>             more fields
>         }
> ```
> 
> Not tested but something like that 😄

We can probably fix this in a later issue, already merged into kk

PR #372, opened 2023-09-27
Stop event month text from wrapping on big screens

MirruK:
Add css rule to stop the responsive font resizing from causing the date start month text (eg. "Oct") from wrapping on larger viewport sizes on events page.


Nannemara (approved):
Beautiful, beautiful

jaspnas (approved):
@Nannemara said it's fine, merging

PR #376, opened 2023-10-06
Fix tests into working condition

MirruK:
Fix all tests so that they don't give random errors and so that they pass for the current state of origin/develop

Add unit tests to CI pipeline

Change the date-test command so that you can provide an optional argument that specifies which test module to run. 
Ex. `date-test news` will run only tests found in the "news" module.

### DIFF
--- a/archive/tests.py
+++ b/archive/tests.py
@@ -4,44 +4,58 @@ from django.conf import settings
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase
 from django.utils import timezone
+from PIL import Image
 
 from archive.models import TYPE_CHOICES, Collection, Document, Picture
 
 
+def create_collection(title="Test collection", collection_type=None):
+    return Collection.objects.create(title=title, pub_date=timezone.now(), type=collection_type)
+
+
+def create_picture(favorite=False):
+    collection = create_collection(collection_type=TYPE_CHOICES[0][1])
+    img = Image.new('RGB', (100, 100))
+    img.save(os.path.join(settings.MEDIA_ROOT, 'test_image.jpg'))
+    img.close()
+    img_file = open(os.path.join(settings.MEDIA_ROOT, 'test_image.jpg'),"rb")
+    img_data = img_file.read()
+    img_file.close()
+    test_image = SimpleUploadedFile(name='test_image.jpg',
+                                    content=img_data,
+                                    content_type='image/jpg')
+    return Picture.objects.create(collection=collection, image=test_image, favorite=favorite)
+
+
+def create_document(title="Test document"):
+    collection = create_collection(collection_type=TYPE_CHOICES[1][1])
+    doc_file = open(os.path.join(settings.MEDIA_ROOT, 'test_document.doc'), "rb")
+    doc_data = doc_file.read()
+    doc_file.close()
+    test_document = SimpleUploadedFile(name='test_document.doc',
+                                        content=doc_data)
+    return Document.objects.create(collection=collection, title=title, document=test_document)
+
+
 # models tests
 class CollectionTestCase(TestCase):
-
-    def create_collection(self, title="Test collection", collection_type=None):
-        return Collection.objects.create(title=title, pub_date=timezone.now(), type=collection_type)
-
     def test_collection_creation(self):
-        c = self.create_collection()
+        c = create_collection(collection_type=TYPE_CHOICES[0][1])
         self.assertTrue(isinstance(c, Collection))
         self.assertEqual(c.__str__(), c.title)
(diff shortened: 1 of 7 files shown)


andreedd:
I wouldn't really restrict the signup for unpublished events, we have a lot of more private events which are unpublished events. For example this years gulissitz, which only people with the link can sign up to.

events/views.py
```diff
@@ -127,6 +135,9 @@ def form_invalid(self, form):
 def ws_send(request, form, public_info):
     ws_schema = 'ws' if settings.DEVELOP else 'wss'
     url = request.META.get('HTTP_HOST')
+    # This indicates that the request is being ran by the tests
+    if not url:
```
andreedd: We could also put something like this: `TESTING = sys.argv[1:2] == ['test']` into core/settings.py and do `if TESTING:` wherever testing specific code is run.

MirruK: Reverted the incorrect unpublished event check and added the suggested TEST environment variable

archive/tests.py
```diff
@@ -4,44 +4,58 @@
...
+                                    content=img_data,
+                                    content_type='image/jpg')
+    return Picture.objects.create(collection=collection, image=test_image, favorite=favorite)
+
 
-    def create_collection(self, title="Test collection", collection_type=None):
-        return Collection.objects.create(title=title, pub_date=timezone.now(), type=collection_type)
+def create_document(title="Test document"):
+    collection = create_collection(collection_type=TYPE_CHOICES[1][1])
+    doc_file = open(os.path.join(settings.MEDIA_ROOT, 'test_document.doc'), "rb")
```
andreedd: This tests seems to fail if we do not have the file.

Chat GPT gave me this nice code of using temp files which seems to work:

```
import tempfile


def create_document(title="Test document"):
    collection = create_collection(collection_type=TYPE_CHOICES[1][1])

    # Create a temporary file with some test data
    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
        temp_file.write(b'This is some test data for the document')
        temp_file.seek(0)  # Reset the file pointer to the beginning
        test_document = SimpleUploadedFile(name='test_document.doc', content=temp_file.read())

    return Document.objects.create(collection=collection, title=title, document=test_document)
```

MirruK: True, that's a nice solution since fixing this with just the "open" function is unnecessarily messy. I swapped out the code and it seems to be working now.

events/tests.py
```diff
@@ -78,24 +86,14 @@ def test_get_events_index(self):
...
                                      sign_up_deadline=timezone.now(),
                                      author_id=self.member.id)
         self.assertIsNotNone(event)
-        c = Client()
+        c = Client(enforce_csrf_checks=False)
         response = c.post(reverse('events:detail', args=[event.slug]), {'user': 'person6', 'email': 'person6@test.com'})
         self.assertEqual(response.status_code, 403)
         self.assertEqual(event.get_registrations().count(), 0)
+
+    # TODO: Test full event
```
jaspnas: TODO

events/tests.py
```diff
@@ -9,8 +9,15 @@
...
-        self.event = Event.objects.create(title='Test event', slug='test', author_id=self.member.id)
+        self.event = Event.objects.create(title='Test event',
+                                          slug='test',
+                                          author_id=self.member.id,
+                                          sign_up_deadline=(timezone.now()-timezone.timedelta(-1))
+                                          )
         self.assertIsNotNone(self.event)
+        print(f"Is open others: {self.event.registration_is_open_others()}")
+        print(f"Is open others time: {self.event.sign_up_others}")
+        print(f"Sign up deadline: {self.event.sign_up_deadline}")
```
jaspnas: Are these needed, it already spams the log enough.

news/tests.py
```diff
@@ -23,11 +23,12 @@ def test_get_news_detail(self):
...
-        response = c.get(reverse('news:detail', args=[post.slug]))
-        self.assertEqual(response.status_code, 200)
-        self.assertIsNone(response.context.get('articles', None))
+    # Get not defined for news view, Client.get() will return error for unpublished article
+    # def test_unpublished_news(self):
+    #     c = Client(enforce_csrf_checks=False)
+    #     post = Post.objects.create(title='Test news2', slug='test2', author_id=self.member.id, published=False)
+    #     response = c.get(reverse('news:detail', args=[post.slug]))
+    #     self.assertEqual(response.status_code, 404)
+    #     self.assertIsNone(response.context.get('articles', None))
```
jaspnas: Maybe delete and redo later?

jaspnas (approved):
Clean up, otherwise looks good

PR #386, opened 2023-10-10
Feat/Check if user has active subscription when signing up for event

### LINKED ISSUE #281: feat: Members signup for event for active paying members
A request from stysse.

When the members sign-up is set for event, it should affect only members who have an active members-subscription., not all members who can sign in. This way people will actually have to pay their yearly DaTe subscription in order to use this benefit.

This could also maybe be implemented as a an additional field which is "active members sign-up" or something

jaspnas:
Fixes #281


MirruK (approved):
Neat change, looks good

PR #387, opened 2023-10-14
CSS variable renaming

jaspnas:
What could go wrong


MirruK (approved):
What's the worst that could happen? Broken css? merge it

PR #392, opened 2023-10-29
Move joke-api call from server to client

MirruK:
Add script tag in html to make the api call to fetch a joke if request errors out in any way, it will fallback and remove the joke header text


jaspnas (approved):
image

PR #396, opened 2023-11-11
improve polls validation

andreedd:


polls/validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+from members.models import ORDINARY_MEMBER
+
+ERROR_MESSAGES = {
+    'not_logged_in': "Logga in för att rösta.",
+    'already_voted': "Du har redan röstat.",
+    'no_choice': "Du valde inget alternativ.",
+    'vote_ended': "Röstandet har avslutats.",
+    'not_authorized': "Du inte är röstberättigad.",
+    'single_choice': "Ändast ett val är tillåtet.",
```
jaspnas: Stav fell

polls/validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if not selected_choices:
+        return render_error(request, question, ERROR_MESSAGES['no_choice'])
+
+    if single_choice_multiple_selected(request, question, selected_choices):
+        return render_error(request, question, ERROR_MESSAGES['single_choice'])
+
+    if not members_question:
+        handle_selected_choices(question, selected_choices, user=None)
+        return HttpResponseRedirect(reverse('polls:results', args=(question.id,)))
```
jaspnas: Does this need to be a separate check or can we rethink this to only have one success at the end of the function?

polls/validation.py
```diff
@@ -0,0 +1,101 @@
...
+    if members_question(question) and not user.is_authenticated:
+        return render_error(request, question, ERROR_MESSAGES['not_logged_in'])
+
+
+def can_member_vote(question, user):
+    if not user.is_authenticated:
+        return False
+
+    if question.members_only:
+        return True
```
jaspnas: Will this not always return true if it's members only without taking other stuff into account

jaspnas:
I can't seem to get voting on public questions working? It tells me I am not allowed to vote?

jaspnas:
It seems like membership restrictions follow the least strict one selected, should this be changed to follow the strictest restriction, or be changed to only allow one type of restriction?

andreedd:
Changed the stupid booleans, wonder who made them

PR #412, opened 2023-11-18
Functionaries

andreedd:
Adds "Funktionärer"

- Members can view,add and remove their functionary roles
- Anyone can view current functionaries
- Members can see and filter functionaries from the past

For further improving DaTe overall documentation.


andreedd:
FunctionaryRoles has to be created through admin first.

core/templates/static/css/base.css
```diff
@@ -62,3 +62,49 @@ div a:hover{
     border-radius: var(--defaultRadius);
 }
 
+/* CUSTOM BUTTON */
+.custom-button {
```
jaspnas: This name should be more descriptive

jaspnas (approved):
Maybe make the custom-button class have more descriptive name, otherwise seems to work

PR #420, opened 2023-11-29
Add random smoke effect

jaspnas:
What were they smoking when the made this one

1 in 420 chance to show a smoke effect when loading any page


JaninaHeikkala:
will this be on date website always from now on?

jaspnas:
As the trigger is currently is set up, yes

If you have better ideas for trigger condition let me know

JaninaHeikkala:
I'm thinking the smoke could appear only in this month and in the future in april (because 4.20 is in april (20th of april)). Or just keep it for a while for now :)

JaninaHeikkala (approved):
Looks good 420

PR #425, opened 2023-12-08
fix error generated from non logged in users

andreedd:


jaspnas (approved):
Looks good

Seal of approval

PR #433, opened 2023-12-30
feat: Celery mail

jaspnas:
Send email using celery workers


core/utils.py
```diff
@@ -28,4 +31,12 @@ def validate_captcha(response: str) -> bool:
...
\ No newline at end of file
+    return res.json().get('success', False)
+
+
+@shared_task
+def send_email_task(*args, **kwargs) -> None:
+    try:
+        send_mail(*args, **kwargs)
+    except SMTPException:
+        print(f"Failed sending email to: {args[3] or kwargs.get('to', '')}")
```
theoabw: Might be better to use logger.error() instead?

members/templates/registration/password_reset_email.html
```diff
@@ -4,11 +4,11 @@
 
 {{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}
```
theoabw, line 5: Might wanna try `<a href="{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}">{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}</a>` to ensure clickability?

Theoabw (changes requested):

core/utils.py
jaspnas: Good catch

members/templates/registration/password_reset_email.html
jaspnas: Not going to change it in this PR, as it seems to work for now without it.

Theoabw (approved):

PR #435, opened 2024-01-01
Return 404 instead of 500 if staticpage does not exist, clean staticp…

### LINKED ISSUE #422: bug: Staticpages return 500 if page is not found
Should return 404 instead if it's not found.

jaspnas:
…ages module

closes #422


andreedd (approved):
Should be goood

PR #449, opened 2024-01-15
Fix/365 fix replace fontawesome kits

### LINKED ISSUE #365: fix: Replace fontawesome kits
Fontawesome kits have a quite restrictive amount of free pageviews (10k), and as the next tier costs 99€ we should be looking into alternatives before this becomes a huge problem. 

Alternatives:
- bootstrap
- google
- ...

Theoabw:
Line awesome should work as a drop in replacement for font awesome and it doesn't have a pageview limit like font awesome.

Closes #365


jaspnas (approved):
Looks good

PR #450, opened 2024-01-15
Add caching to rarely updated pages

jaspnas:
5 min TTL for main page

1 minute TTL for index- and staticpages

Makes the website query less data from the database


Theoabw (approved):
lgtm

PR #453, opened 2024-01-16
Fix/280 fix clean up the spaghetti in eventsviewspy

### LINKED ISSUE #280: fix: Clean up the spaghetti in events/views.py
Big pasta in there!

We need to have a proper Tagliatelle night to fix it.

Theoabw:
Check `avec_data = form.cleaned_data.copy()` and if it needs deepcopy or not. The functionality should basically be the same otherwise.

Closes #280

### DIFF
--- a/events/tests.py
+++ b/events/tests.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.models import User
 from django.test import Client, TestCase
 from django.urls import reverse
 from django.utils import timezone
@@ -10,23 +11,28 @@ class EventTestCase(TestCase):
     def setUp(self):
         self.member = Member.objects.create(username='Test', password='test', is_superuser=True)
 
+        now = timezone.now()
+        yesterday = now - timezone.timedelta(days=1)
+        two_days_ago = now - timezone.timedelta(days=2)
+        tomorrow = now + timezone.timedelta(days=1)
+
         # Create past and future events for testing
         self.past_event1 = Event.objects.create(
             title='Past Event 1',
             slug='past-event-1',
-            event_date_end=timezone.now() - timezone.timedelta(days=2),
+            event_date_end=two_days_ago,
             author_id=self.member.id
         )
         self.past_event2 = Event.objects.create(
             title='Past Event 2',
             slug='past-event-2',
-            event_date_end=timezone.now() - timezone.timedelta(days=1),
+            event_date_end=yesterday,
             author_id=self.member.id
         )
         self.future_event = Event.objects.create(
             title='Future Event',
             slug='future-event',
-            event_date_end=timezone.now() + timezone.timedelta(days=1),
+            event_date_end=tomorrow,
             author_id=self.member.id
         )
         self.event = Event.objects.create(title='Test event',
--- a/events/tests.py
+++ b/events/tests.py
@@ -1,8 +1,11 @@
-from django.test import Client, TestCase
+import datetime
+from django.test import Client, TestCase, RequestFactory
 from django.urls import reverse
 from django.utils import timezone
 
+from events.forms import PasscodeForm
 from events.models import Event
+from events.views import EventDetailView
 from members.models import Member, ORDINARY_MEMBER, Subscription, SubscriptionPayment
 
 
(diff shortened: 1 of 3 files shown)


andreedd:
Would probably be good at some point to split up the validation in some kind of validation file. A bit like what was done on polls.

theoabw:
> he functionality should basically be the same otherwise.

I totally didn't say that. Anyway, tests have passed after wayy too many tries. My fleshed out test member in tests.py is kind of unnecessary for this pr but might be useful in the future. I've tested it on localhost and basic event signup seems to be working fine. Logs don't give any weird errors at least.

theoabw:
I still don't quite know how

```
if self.object.sign_up and (request.user.is_authenticated
                                    and self.object.registration_is_open_members()
                                    and Member.objects.get(username=request.user.username).get_active_subscription() is not None
                                    or self.object.registration_is_open_others()
                                    or request.user.groups.filter(
                    name="commodore").exists()):
```
                    
In the original code didn't throw an exception and my refactored version that defined the bools beforehand, but kept the if statement the same, did?

events/views.py
```diff
@@ -58,119 +58,120 @@ def get_context_data(self, **kwargs):
...
-        if redirect_link and show_content:
-            return redirect(redirect_link)
+    def get_template_names(self):
+        event_title = self.get_context_data().get('event').title.lower()
+        logger.debug(event_title)
+        if event_title in ['årsfest', 'årsfest gäster']:
+            return ['events/arsfest.html']
+        if self.object.passcode and self.object.passcode != self.request.session.get('passcode_status', False):
+            return ['events/event_passcode.html']
+        return [self.template_name]
```
jaspnas, line 76: This wasn't a list previously, is this correct?

events/tests.py
```diff
@@ -3,12 +3,41 @@
...
+
+        subscription = Subscription.objects.create(
+            name="Basic Subscription",
+            does_expire=True,
+            renewal_scale="year",
+            renewal_period=1,
+            price=100.00
+        )
+
+        SubscriptionPayment.objects.create(
```
jaspnas: Would be nice to have this as a variable for the class as well to be able to adjust the subscription status

events/tests.py
```diff
@@ -3,12 +3,41 @@
...
+            email="testuser@example.com",
+            first_name="Test",
+            last_name="User",
+            phone="123456789",
+            address="123 Test Street",
+            zip_code="00000",
+            city="Test City",
+            country="Finland",
+            membership_type=ORDINARY_MEMBER,  # Using one of your defined membership types
+            is_superuser=True
```
jaspnas: Set this to false, superusers should not sign up for events in production

jaspnas (dismissed):
Resolve the comments and rebase, looks like it's working for most normal use cases

jaspnas (dismissed):
Needs more tests, should at least test the different types of members with different subscription payment statuses.

theoabw:
> Needs more tests, should at least test the different types of members with different subscription payment statuses.

> Would probably be good at some point to split up the validation in some kind of validation file. A bit like what was done on polls.

I'm working on splitting up validation into a separate validation file. I noticed that the detail template was only checking for if user was validated and not if the user had an active subscription. This leads to members without an active subscription being able to see the registration form, but not being able to register to the event if self.object.registration_is_open_members() is true. 

Members without an active subscription shouldn't be able to see the registration form for an event that's only open for members.

theoabw:
1. TODO: Captcha test
2. TODO: avec test
3. TODO: redirect link test
4. TODO: anonymous test
5. TODO: custom field test
6. TODO: max participants test

events/tests.py
```diff
@@ -87,11 +135,134 @@ def test_get_events_index(self):
...
-        self.assertEqual(event.get_registrations().count(), 0)
+        self.assertEqual(self.event.get_registrations().count(), 0)
+
+    def test_event_detail_view_with_and_without_passcode(self):
+        self.event.passcode = 'secret'
+        self.event.save()
+        c = Client()
+        wrong, right = {'passcode': 'wrong'}, {'passcode': 'secret'}
+        incorrect = c.post(reverse('events:detail', args=[self.event.slug]), wrong)
+        self.assertEqual(incorrect.status_code, 200)
```
jaspnas, line 152: Should this be changed to return 401

Can be done as a later issue (create and resolve this comment)

jaspnas (approved):

PR #458, opened 2024-01-17
Update dependencies

### LINKED ISSUE #424: Update dependencies
Dependencies should be updated to latest versions, this includes:

- Django 5.0
- Postgres 16
- Python 3.12

+ other smaller dependencies

Theoabw:
from django.utils.timezone import utc

becomes

import datetime

utc = datetime.timezone.utc


events/migrations/0001_initial.py
```diff
@@ -8,9 +7,9 @@
 import django.utils.timezone
 from django.conf import settings
 from django.db import migrations, models
-from django.utils.timezone import utc
 
 
+utc = datetime.timezone.utc
```
jaspnas, line 12: Need to test on QA if this will break stuff (should be fine as functionality should be the same)

jaspnas (approved):
Looks good, please hold merging this until someone has time to update QA

PR #460, opened 2024-01-19
Isolated Albin

### LINKED ISSUE #451: fix: load SVGs from files
SVGs are currently straight in the code.

Move SVGs into files and load them from there.

**Additional context**  
Check start.html for example

Theoabw:
Made start.html cleaner by using Django's include instead of having Albin's SVG there. Inspired by date_25.html


theoabw:
Someone else can probably find a better way of doing this. I'm running into problems with the css and animations when I try to reference it as an svg file in an object, iframe, or img tag.

jaspnas:
Would be nice not having this rendered into html, but loaded as an image/static resource

jaspnas (approved):

PR #465, opened 2024-01-24
Calendar retry

jaspnas:
Probably has weird bugs from rebasing left


date/static/date/css/homepage.css
```diff
@@ -434,9 +434,10 @@ div.news-content a:hover{
     grid-template-areas: 
     ". box ."
     ". calendar .";
-    padding-bottom: 10px;
 }
-
+#calendar{
+    margin-bottom: 1rem;
```
theoabw, line 439: Using font size of root element as margin size? Might be better with vh.

date/views.py
```diff
@@ -28,18 +24,31 @@ def index(request):
     if aa_posts and aa_posts[0].published_time > time_since:
         aa_post = aa_posts[0]
 
+    def calendar_format(x):
+        formatstr = "%H:%M"
+        calendar_events_dict = {}
+        for e in x:
+            event_url = "events/" + e.slug
+            event_dict = {e.event_date_start.strftime("%Y-%m-%d") :
```
theoabw, line 32: Make it possible to see past events?

MirruK (approved):
This looks to be in working condition, feature and styling changes can be added later if needed.

PR #476, opened 2024-01-28
Calendar shows old events (max 1 month)

MirruK:
Change how the timestamp for the event
is rendered so that it accounts for client timezone

Fix variable naming conflict


jaspnas (approved):
lgtm

PR #497, opened 2024-02-04
Fix/form resubmit on refresh

### LINKED ISSUE #496: fix: Form resubmit on page refresh
When refreshing the page after submitting a form, the browser asks the user if they want to resubmit the form. Usually users don't intend to resubmit data with a page refresh.

Pages affected:
- Event pages
- Social harassment form
- Member signup
- CTF pages

Do a redirect after submitting the form

**Additional context**  
https://en.wikipedia.org/wiki/Post/Redirect/Get

keanderss:
Closes #496

### DIFF
--- a/ctf/views.py
+++ b/ctf/views.py
@@ -1,4 +1,4 @@
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.views import generic
 from django.http import HttpResponseForbidden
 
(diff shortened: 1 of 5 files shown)


keanderss:
Event test fail because they get a 302 response instead of 200 due to the redirect.

jaspnas:
Needs changes to the tests to 302, and preferrably an assertion to check that the redirect target is correct

ctf/views.py
```diff
@@ -59,16 +66,18 @@ def flag(request, ctf_slug, flag_slug):
...
                     if flag.exists():
+                        request.session['flag_valid'] = True
                         if user_solved or flag.first().solver:
                             # User has already solved a flag or flag is already solved
-                            return form_valid(request, context)
+                            return redirect(request.path)
                         flag.update(solver=request.user, solved_date=datetime.datetime.now())
                         logger.info(f'Solver: {flag.first().solver}')
                         context['flag'] = flag.first()
                         context['solved'] = True
```
jaspnas, line 76: Is this needed anymore?

jaspnas (approved):
Looks good

ctf/views.py
keanderss: Seems to be tied to the logic of how the page is rendered. It doesn't seem to affect any functionality since the page is reloaded anyways.

PR #502, opened 2024-02-06
ctf logs in admin

Theoabw:
tested and lgtm at least.


jaspnas:
This needs the ability to filter based on CTF and show the CTF,  I would prefer if this would be an inline like the event signups under the CTF itself

jaspnas (changes requested):
Needs improvement to filtering(and maybe the admin view), otherwise looks fine

date/static/date/css/homepage.css
```diff
@@ -509,7 +509,7 @@ div.calendar-container a {
    * Dont use this class for anything else*/
   color: black;
   font-size: 0.75rem;
-}
+}git
```
jaspnas: This should not be here

jaspnas (changes requested):

jaspnas (approved):
lgtm

Issue #348, opened 2023-07-03
Clean up settings


jaspnas:
The django settings file should be cleaned up and organized if we want to have different settings files for each websites

We should also identify what would be common config and what is site-specific.

PR #524, opened 2024-03-24
Prod rework

jaspnas:


docker-compose.prod.yml
```diff
@@ -0,0 +1,103 @@
...
+    volumes:
+        - ./nginx.conf:/etc/nginx/conf.d/default.conf
+
+volumes:
+  date_postgres_data:
+
+#networks: 
+#  default: 
+#    external: 
+#      name: external-example
```
theoabw, line 103: Does this need to be here?

docker-compose.prod.yml
```diff
@@ -0,0 +1,103 @@
...
+      - ALLOWED_HOSTS=${ALLOWED_HOSTS}
+      - ALLOWED_ORIGINS=${ALLOWED_ORIGINS}
+      - EXTRA_STAFF_GROUPS=${EXTRA_STAFF_GROUPS}
+      - EMAIL_HOST_RECEIVER=${EMAIL_HOST_RECEIVER}
+      - DEFAULT_FROM_EMAIL=${DEFAULT_FROM_EMAIL}
+      - EMAIL_HOST=${EMAIL_HOST}
+      - EMAIL_HOST_USER=${EMAIL_HOST_USER}
+      - EMAIL_HOST_PASSWORD=${EMAIL_HOST_PASSWORD}
+      - CF_TURNSTILE_SITE_KEY=${CF_TURNSTILE_SITE_KEY}
+      - CF_TURNSTILE_SECRET_KEY=${CF_TURNSTILE_SECRET_KEY}
```
theoabw, line 38: Consider using env_file directive with separate env files for each container if needed. Could make the compose files cleaner. "web" and "asgi" seems to have the same env vars defined so they could maybe have a common env file that's defined with the env_file directive.

jaspnas: Topic for a later issue, problem is some variable names are different for host and container and all env variables are currently sourced into host env first. 

Probably should be tackled in #348

docker-compose.prod.yml
andreedd: Nope, I think this was some old network conf was from when I tried some Minio stuff :D

andreedd (commented):
Looks pretty clean overall!

And yep would probably be a good idea to clean up env variables to some .env file if still continuing with compose

Theoabw (approved):
lgtm as long as the clutter get cleaned up at some point. You could add comments about this to #348 and merge.

jaspnas:
> lgtm as long as the clutter get cleaned up at some point. You could add comments about this to #348 and merge.

Created #528 and #529 to address this

docker-compose.prod.yml
andreedd: We could also start using something like doppler for secrets.
https://www.doppler.com/

andreedd: Free subscription for students :)
All the env variables should also have default values in the app, e.g. `.get(EMAIL_HOST_USER, 'admin@example.com')`

jaspnas: Tracked in #531 . Will probably be done later to limit scope of project

PR #526, opened 2024-03-25
Add translations for static & dynamic content

MirruK:
Allow translating of navbar and event title/content.

TODO: translate all the static content to English and Finnish

TODO: Add fallbacks for event title and content when no translations are provided


README.md
jaspnas, line 1: Probably best to remove all new migrations and re-create them from scratch,

andreedd: Yep, only 1 migration file per PR per app is necessary

PR #540, opened 2024-03-31
Backend merge: Project restructure

### LINKED ISSUE #335: Deploy modifying only one file + env
There is currently a need to modify at least `core/.env` from the state that it is in in git to get a working production server. 

On the other sites `docker-compose.yml` file is also modified. The things that need to be changed in these files should be added to the environment variables or in some other way that doesn't require separate configs in production.

### LINKED ISSUE #338: Separate common and site specific frontend files
Make directories for templates and  static, inside those separate comon and site specifc (only DaTe to begin with) files into separate subfolders - common and date.

Make sure django can still read both the common and site-specific files.

### LINKED ISSUE #348: Clean up settings

### LINKED ISSUE #542: feat: Replace AA news hard-code
**Is your feature request related to a problem? Please describe.**
Currently AA news is hardcoded as a separate model.

Maybe consider adding support for dynamic categories and have a default category that will show up on the homepage.

**Additional context**
Might be hard with migrations

### LINKED ISSUE #544: fix: One table for static page navs
Current implementation of static page navs is clunky with 2 tables for staticpages and urls

Only have 1 table for everything allowing you to freely reorder

**Additional context**  
Should have reorderable admin

IMO it's fine if we drop the static page special sauce and just have e.g. a lookup for if it's a static page as a backup.

jaspnas:
Adding support for running multiple websites from the same branch


core/settings/common.py
```diff
@@ -0,0 +1,280 @@
...
+# https://docs.djangoproject.com/en/2.1/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'postgres',
+        'USER': 'postgres',
+        'PASSWORD': env('DB_PASSWORD', default=''),
+        'HOST': 'db',
+        'PORT': 5432,
```
andreedd: These could maybe be env vars?

andreedd (commented):
Daamn, nicee :D

core/settings/common.py
jaspnas: Still need to go over this several times, so very much WIP for now :D

andreedd (commented):
I like this solution. Good job! Did not test, just read changes.

Probably worth checking the css var colors still at some point, either here or in a new issue. Might be that just changing the css root vars is not enough. I think I also did some other small changes for the other sites, not sure.

jaspnas:
> Probably worth checking the css var colors still at some point, either here or in a new issue. Might be that just changing the css root vars is not enough. I think I also did some other small changes for the other sites, not sure.

Will be a problem for once we throw in frontend from other sites into the main branch. Best to limit the scope now so we get this merged

jaspnas:
FYI for anyone reviewing, some statics may have the wrong path as the path has changed in django but not in the templates, please add a comment containing which static file doesnt load and which page it occurs on if you find anything

jaspnas:
Should now have all the functionality changes fully complete.

theoabw:
The style for the members/info table isn't dependent on the project css vars. Should be fixed at some point.

Theoabw (approved):
lgtm

Issue #330, opened 2023-06-02
feat: Unified backend


jaspnas:
Merge conflicts is becoming an increasingly bad problem when things need to be updated across all sites

There doesn't seem to be any need for having different backends for the different websites as most things are similar enough that differences should be able to be controlled by simple environment variables.

### Initial idea

- Move all sites frontend files to the main develop branch and control which one is used based on environment variable
- Add the ability to enable and disable modules using environment variables. 
- Make any custom functionality be enabled by environment variables.
- MIgrate the different databases to use the same models

andreedd:
Yes

andreedd:
Se should create a mew Django app for both KK and Biocum. In those folders those all the association custom stuff should be located.

jaspnas:
> Se should create a mew Django app for both KK and Biocum. In those folders those all the association custom stuff should be located.

Is there actually anything major enough that it would need it's own apps. You can probably get away with just setting the disabled modules in .env to start with. Or setting some extra modules variable.

Then for minor stuff you can just check the env variable, but I think you can remove most of the differences on the backend side.

Templates/folders will probably be moved under root and then have folders for common, date, kk etc.

andreedd:
I think there will be from time to time that the projects require custom solutions. I think it would be better to keep all of those custom solutions in the same place. For example the custom events templates could be moved there. I know KK has an lucia app as well, which also could be moved there. DaTe has the CTF app, do we want that one as DaTe only? And the list goes on..

But of course these things could also be where they are now and just be activated using env variables.

jaspnas:
I feel like if we have all custom solutions for each project in only one module it will become a massive module that will just keep growing, especially for DaTe where we have most potential for custom features. 

Doing more major backend code changes than we need to will also increase the scope of this project that is already massive for DaTe standards. 

Also if more than one, but not all websites needs the custom functionality we would essentially have to duplicate the code if it's in the one custom module. 

As for custom event templates we should probably figure out a better way to do it than just hard-coding or just having the option to specify a pre-hardcoded template in the model.

I think it would still make the most sense to just be able to enable and disable the modules, or have some kind of launch-configs file that specifies launch config based on what site we're hosting.

jaspnas:
We should open up smaller issues for things that aren't totally all-encompassing. Otherwise this issue will become a total mess.

andreedd:
Yeah true, the hardcoded template shouldn't really be needed if it is done right :D

[...]

PR #546, opened 2024-04-01
Create news categories

jaspnas:
Part 1 of #542 

Needed as part of #330


news/admin.py
```diff
@@ -1,12 +1,17 @@
...
+
+class CategoryAdmin(admin.ModelAdmin):
+    list_display = ('name',)
+    search_fields = ('name',)
 
 
 class PostAdmin(admin.ModelAdmin):
 
-    list_display = ('title', 'author', 'created_time', 'modified_time', 'published', 'albins_angels')
+    list_display = ('title', 'author', 'category', 'created_time', 'modified_time', 'published', 'albins_angels')
```
theoabw, line 14: Albin's angles will have it's own logic that checks if the category is Albin's angels in the future, right?

Theoabw (approved):
Looks good for part 1

news/admin.py
jaspnas: Yes, I just want to deploy in 2 parts so that all the data is in the correct format once we do the frontend change

PR #604, opened 2024-04-25
Make memberpass show login page for logged out users

keanderss:


members/views.py
```diff
@@ -54,6 +54,7 @@ def post(self, request):
 
 
 class CertificateView(View):
+    @method_decorator(login_required)
```
jaspnas, line 57: nice

jaspnas (approved):

PR #630, opened 2024-08-20
updated STORAGES setting

Theoabw:


core/settings/common.py
```diff
@@ -15,6 +15,7 @@
 import json
 
 import environ
+from django.conf.global_settings import STORAGES
```
jaspnas: We redefine this later, do we need it?

jaspnas (approved):

PR #643, opened 2024-10-03
Improve the live attendee list

Mixu78:
This PR fixes an XSS vulnerability in the live attendee list caused by lack of input sanitization, and makes the created table row match the event details template more closely. This is done by translating and adding `<i>` elements to the names of anonymous attendees as well as ensuring fields are output in the same order as in the template.


jaspnas (approved):
Looks like it works as intended and shouldn't cause any problems.

Issue #667, opened 2024-12-13
bug: KK (and maybe DaTe) 500 error in polls.


jaspnas:
**Describe the bug**  

500 error and stacktrace in console probably related to user not being authenticated properly/token being expired.

**To Reproduce**  

Unknown at time of writing.

**Additional context**  

```kkwebsite-web-6     | ERROR Internal Server Error: /polls/66/vote/
kkwebsite-web-6     | Traceback (most recent call last):
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/asgiref/sync.py", line 518, in thread_handler
kkwebsite-web-6     |     raise exc_info[1]
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 42, in inner
kkwebsite-web-6     |     response = await get_response(request)
kkwebsite-web-6     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/asgiref/sync.py", line 518, in thread_handler
kkwebsite-web-6     |     raise exc_info[1]
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/django/core/handlers/base.py", line 253, in _get_response_async
kkwebsite-web-6     |     response = await wrapped_callback(
kkwebsite-web-6     |                ^^^^^^^^^^^^^^^^^^^^^^^
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/asgiref/sync.py", line 468, in __call__
kkwebsite-web-6     |     ret = await asyncio.shield(exec_coro)
kkwebsite-web-6     |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/asgiref/current_thread_executor.py", line 40, in run
kkwebsite-web-6     |     result = self.fn(*self.args, **self.kwargs)
kkwebsite-web-6     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
kkwebsite-web-6     |   File "/usr/local/lib/python3.12/site-packages/asgiref/sync.py", line 522, in thread_handler
kkwebsite-web-6     |     return func(*args, **kwargs)
kkwebsite-web-6     |            ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #668, opened 2024-12-13
Fix 500 error in polls if user isn't authenticated

jaspnas:
Related to #667


MirruK (approved):
Yep, due to if statements short circuiting, the code will stop at user.is_authenticated, unless they are actually logged in LGTM

PR #678, opened 2024-12-21
Rotate gunicorn workers after a set time

jaspnas:
This should allow us to save some memory without a major impact on performance


keanderss (approved):
saving memory sounds good

PR #680, opened 2024-12-21
KK UI bugs

### LINKED ISSUE #675: bug: KK post migration visual bugs
**Describe the bug**  

Some colors and buttons are wrong on KK after migration

These are at least present in archive and members, at least views with a table. 

**Screenshots**  

image

keanderss:
Closes #675


templates/common/members/functionary.html
```diff
@@ -11,7 +11,7 @@
             <div class="content">
                 <!-- Functionary Table Section -->
                 <h4>Funktionärsposter</h4>
-                <table class="table table-dark table-striped">
```
jaspnas, line 14: KK isn't using functionary yet, also will this not break DaTe?

jaspnas (changes requested):
Maybe we should create a default-table CSS class that we use and can define per site

templates/common/members/functionary.html
keanderss: Yes, on closer inspection this change would break DaTe I just forgot to check where i was making changes.

templates/common/members/userinfo.html
```diff
@@ -20,7 +20,7 @@
                 <!-- User Information Section -->
                 <h3>{{ user.username }}</h3>
                 <div id="userInfo">
-                    <table class="table table-dark table-striped">
+                    <table id="table" class="table table-striped">
```
keanderss, line 23: Would have preferred to use classes for this but for some reason they were not being applied to DaTe.

keanderss:
TODO:
- Fix mobile calendar 
- Fix mobile sponsors

keanderss:
For some reason the old footer was not removed from the files which caused the footer issues.

Summary of changes:
- Removed old footer
- Removed hardcoded sponsors (should probably be with the other sponsors)
- Added calendar and sponsors to mobile view
- Fixed dynamic association logo resizing
- Some other minor changes to make KK more consistent with DaTe

templates/kk/core/footer.html
```diff
@@ -1,54 +0,0 @@
...
-        <div class="col-4">
-            <p>
-                {% trans 'Kemistklubben vid Åbo Akademi rf' %}<br>
-                Aurum
-                <br>
-                Henriksgatan 2
-                <br>
-                20500 Åbo
-                <br>
-                Kanslitid Må - To 11:30-13
```
jaspnas, line 42: KK needs their own footer template because they have their "Kanslitid". Maybe it can be a content variable, but maybe too much work to make the common footer dynamic for this if only KK is using it?

templates/kk/date/start.html
```diff
@@ -58,9 +60,7 @@ <h3>{% trans 'Inga nyheter hittades' %}...</h3>
                 </div>
                 <div class="sidebar-container">
                     <div class="events-container">
-                        <a id="andritz-logo" href="https://www.andritz.com/group-en"><img class="sidebar-logo-container" src={% static "/core/images/andritzLogo.png" %} alt="Andritz"></a>
```
jaspnas, line 61: These were specifically requested by KK in this form, so let's keep them there.

templates/kk/date/start.html
```diff
@@ -151,8 +214,8 @@ <h3 class="mt-3">Samarbetspartners</h3>
             const tempDate = new Date(o[1].eventFullDate);
             let hours = tempDate.getHours();
             let minutes = tempDate.getMinutes();
-            // Ensure minutes are two digits
-            minutes = minutes < 10 ? '0' + minutes : minutes;
+            // Ensure hours and minutes are displayed with two digits
+            hours = hours < 10 ? '0' + hours : hours;
```
jaspnas, line 155: Do we not need another line here for minutes?

templates/date/date/start.html
```diff
@@ -180,7 +180,7 @@ <h3 class="header-border-bottom">Samarbetspartners</h3>
                         <div class="d-flex flex-wrap justify-content-center align-items-center">
                             {%for ad in ads%}
                                 <a class="m-2" href="{{ ad.company_url }}">
-                                    <img src="{{ ad.ad_url }}" alt="img">
+                                    <img src="{{ ad.ad_url }}" alt="img" onerror="this.src='/static/core/images/fallback.png'">
```
jaspnas, line 123: Nice, maybe add this to Biocum as well after migration.

jaspnas (changes requested):

templates/kk/date/start.html
```diff
@@ -117,14 +117,77 @@ <h3 class="mt-3">Samarbetspartners</h3>
...
-                                        <img src="{{ ad.ad_url }}" alt="img">
+                                        <img src="{{ ad.ad_url }}" alt="img" onerror="this.src='/static/core/images/fallback.png'">
                                     </a>
                                 </li>
                                 {% endfor %}
                             </ul>
                         </div>
                     </div>
                 </div>
+                <div class="mobile-events-container">
```
jaspnas: Probably should see if we can fix this without code duplication. Maybe create a new issue to fix it later.

templates/date/date/start.html
jaspnas: Probably also want to add this for the mobile container?

jaspnas (approved):
Looks good

PR #682, opened 2025-01-02
Fix issues with pdf viewer

### LINKED ISSUE #633: fix: publications last page back not working correctly
**Problem**

When navigating one page backwards from the last page it goes back only 1 page and shows the second last page and the back cover.

The navigation should skip to the n - 2 pages from the last page. the relevant js can be found from static/common/publications/js (probably in event listeners)

### LINKED ISSUE #634: fix: zoom interaction in publications
**Problem**

The arrow keys control the zoom percentage too unless the user clicks elsewhere on the site after selecting a percentage. This is inconvenient if one chooses to use the arrow keys for navigation.

**Solution**

The arrow keys shouldn't interact with the zoom percentage after the user has made a selection.

joandste:
Closes #633, Closes #634

### DIFF
--- a/static/common/publications/js/eventListeners.js
+++ b/static/common/publications/js/eventListeners.js
@@ -51,6 +51,12 @@ export function setupEventListeners(state) {
     });
 
     document.getElementById('zoom-select').addEventListener('change', (e) => {
+        // Disable tag after its been used
+        document.getElementById('zoom-select').disabled = true;
+        // Re-enable it after a short delay
+        setTimeout(() => {
+            document.getElementById('zoom-select').disabled = false;
+        }, 100);
         updateState({ scale: parseFloat(e.target.value) });
         renderPages(state);
     });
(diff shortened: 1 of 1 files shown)


joandste:
Zoom might be quite difficult to fix since the `<select>` tag after having been interacted with reacts with all the arrow keys until you click away.
One solution might be having multiple buttons for the zoom levels instead of a drop down menu.

joandste:
Disabling the tag temporarily after its been used seemed to do the trick.

keanderss:
Currently when viewing a document that has an _**even**_ page count the page ordering isn't retained when going backwards and the second page won't be displayed at all. For documents with an _**odd**_ page count everything works as intended. The changes make it so that this behaviour is now flipped and will break documents with an _**odd**_ page count in the same way while fixing it for documents with an _**even**_ page count.

joandste:
Yeah sorry, i just assumed we were going to always use even pages, because then it will always be possible to show the front and backcover as a single page

Im guessing we want to show a double view on the last page if document has an odd number of pages?

Though books cant have an odd number of pages.

keanderss:
For the current use case of the publications app it won't be a problem as it is only used for UpDaTe but this could change in the future.

Theoabw (approved):
Works, lgtm

PR #683, opened 2025-01-03
Redirect url directly

### LINKED ISSUE #562: bug: External event redirects not opening in new tab
**Describe the bug**  

Event redirects to external site doesn't open a new tab. Caused by index linking to https://

**Additional context**  

Make event index link directly to the external site

joandste:
Closes #562, Uses redirect_url directly in template if field is non nil

### DIFF
--- a/events/views.py
+++ b/events/views.py
@@ -36,6 +36,8 @@ class EventDetailView(DetailView):
 
     def get(self, request, *args, **kwargs):
         self.object = self.get_object()
+        if self.object.redirect_link:
+            return HttpResponseForbidden()
         context = self.get_context_data(object=self.object)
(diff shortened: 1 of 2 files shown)


joandste:
Probably have to remove some now unneeded backend stuff

joandste:
Link has to be hidden in some cases aswell

joandste:
Worth mentioning that the tests fail because the old way of redirecting (EventDetailView.get()) returns status 403 now when redirecting is used

jaspnas:
I think we still want to return a redirect, to not break existing links that are expected to redirect elsewhere. That would still be better than a 403.

Theoabw (approved):
lgtm

PR #687, opened 2025-01-11
Change dev env restart policy to unless-stopped

keanderss:
Currently dev containers start automatically whenever docker starts. Change this behavior so they only start when the user starts them.


jaspnas (approved):
Looks good, shouldn't cause problems now that we have separate prod config

PR #691, opened 2025-01-12
Set celery concurrency to 2 to preserve memory

jaspnas:
Currently celery starts as many tasks as there is CPU cores, each task uses memory so we can save quite a lot by starting less tasks. 

Celery currently has almost no traffic and none of it is long running or time-critical. May need to be looked at in the future, but for now 2 should be adequate.


jaspnas:
On my machine memory usage went down from ~1.65GiB to ~150MiB. So pretty big gains especially when spread out across all our deployments

PR #718, opened 2025-03-09
BIOCUM: Change to using DaTe models

jaspnas:
Admin files/forms are only changed to resolve errors

This is not supposed to run in production, it's just a migration step to get the data into a compatible format


Theoabw (approved):
lgtm

PR #733, opened 2025-04-07
Add action to admin panel allowing deletion of participants

MirruK:
Allows easy deletion of attendees from old events

### DIFF
--- a/events/admin.py
+++ b/events/admin.py
@@ -1,8 +1,11 @@
 import logging
+from django.http import HttpResponseRedirect
+from django.contrib import messages
 
 from admin_ordering.admin import OrderableAdmin
 from django.contrib import admin
 from django.db.models import JSONField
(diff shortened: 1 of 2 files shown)


Theoabw (approved):
works + lgtm

events/admin.py
```diff
@@ -92,6 +98,35 @@ def account_actions(self, obj):
...
+
+        # Collect all attendees to be deleted
+        attendees_to_delete = []
+        for event in queryset:
+            attendees_to_delete.extend(event.eventattendees_set.all())
+
+        if 'confirm' in request.POST:
+            # If the user confirms, delete all attendees
+            for attendee in attendees_to_delete:
+                attendee.delete()
```
jaspnas, line 114: Do we want to delete fully, or just anonymise and delete custom preferences

Issue #528, opened 2024-03-26
feat: Replace date-* commands with pyinvoke or similar


jaspnas:
The date-* commands cause problems when trying to use e.g. `example.env` as an env file. 

Should replace it with pyinvoke tasks or something similar:
https://www.pyinvoke.org/

theoabw:
Closing as stale: the underlying problem (date-* helpers conflicting with env files) was fixed in #923, `env.sh` now only registers checkout-aware aliases without loading or changing app configuration, compose selection lives in `.env` via `COMPOSE_FILE`, and CI no longer sources env.sh. The pyinvoke migration itself was never done, but the blockers it addressed are gone. Reopen if the alias approach becomes a problem again.

PR #758, opened 2025-09-01
Fix: #528 and #529

### LINKED ISSUE #529: fix: Same env-variable naming on host and container
Currently the naming for environment variables is different on the host and in the container

Rename them so they are the same on both sides and remove the manual mappings in docker compose configs.

keanderss:
Closes: #529 

- Cleaned up compose files:
  - Reduced duplication by making docker-compose.prod.yml work as override for docker-compose.yml
    - https://docs.docker.com/compose/how-tos/production/ 
  - Added database service healthcheck
    - Replaces wait-for-postgres.sh  
  - Uses env files directly, ENV_FILE is defined in master.example.env. 
  - Configured compose watch mode instead of using volumes for syncing code
  - Moved long startup commands into web.prod.sh and web.dev.sh
  - Removed networks since they only seemed to define internal/external networks in name and not in configuration
    - Didn't seem to do anything except create confusion

- Moved aliases and stuff to master.example.env to make example.env readable by compose
  - Used this name since it bypasses .gitignore
  - This is a workaround for #528 for now
  - Added --rm to aliases to stop them from creating orphans every time they are used

- Decoupled cache backend in dev mode

Should look into #528 before merging


theoabw:
TODO for myself: check if any changes here are relevant after the changes that I already did to the composes and aliases etc.

keanderss:
> TODO for myself: check if any changes here are relevant after the changes that I already did to the composes and aliases etc.

I don't really remember since it's been a long time since i looked at this but the changes in commit https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/pull/758/changes/f19bb90b6f63f8e0b67d6597a0106902784130e1 might still be relevant. Basically use a dummy cache instead of using Redis/Valkey so you can more easily work on the templates without having the cache get in your way.

https://docs.djangoproject.com/en/6.0/topics/cache/#dummy-caching-for-development

Theoabw:
dev dummy cache done in #828

PR #783, opened 2025-09-29
Fix ON100 issues

nic0r:


events/views.py
```diff
@@ -169,6 +161,8 @@ def redirect_after_signup(self):
         event = self.get_context_data().get('event')
         if event.title.lower() in ['årsfest', 'årsfest gäster']:
             return redirect(f"{reverse('events:detail', args=[event.slug])}#/anmalda")
+        elif event.title.lower() in ['österbottniska nationens 100-årsjubileum'] or 'ön100' in event.title.lower():
```
jaspnas, line 164: Might be error prone or not catch everything. but shouldn't break anything for now

jaspnas (approved):

PR #794, opened 2025-10-24
Functionality for Membership Request Form (Ansökan om medlemskap)

MirruK:
TODO: Better styling and validation for the form view


MirruK:
TODO: Ask stysse which fields should be required / optional

PR #796, opened 2025-10-27
Biocum frontend compatability

Theoabw:
Added biocum frontend so that it can be merged into develop and be ready for biocum models


theoabw:
image

small visual glitch in lower left corner of svg should probably be fixed at some point

jaspnas (approved):

PR #802, opened 2025-11-04
added k8s manifest

Theoabw:


theoabw:
Just leaving this here in case we decide to change to k8s:

Once the cluster has a running Postgres pod (with its PVC already provisioned), you can load the old postgres dump in a few ways:

  - `kubectl port-forward statefulset/db 5432:5432 then run psql -h 127.0.0.1 -U postgres -d yourdb < dump.sql` from the machine that hosts the dump. This reuses the local tooling you already have in Docker-land; just point it at the forwarded port instead of the Docker service.
  - Copy the dump straight into the pod and restore in-place: `kubectl cp dump.sql date-website/db-0:/tmp/dump.sql` followed by `kubectl exec -it date-website/db-0 -- psql -U postgres -d yourdb -f /tmp/dump.sql`.
  - For very large dumps (or if you need this repeatable), create a temporary Kubernetes Job that mounts the same PVC as the database and runs a restore script (psql < dump.sql). You can bake the dump into a ConfigMap, mount it from a bucket, or have the Job curl it from object storage.

PR #828, opened 2026-02-17
Add dummy cache for development

keanderss:
Adds dummy cache for development to make modifying templates less annoying.


Theoabw (approved):
Nice! lgtm

PR #829, opened 2026-02-20
frontend decoupling

Theoabw:


theoabw:
Still a work in progress. We still don't have good parity with the old legacy frontend

theoabw:
Will do another attempt probably after the current codebase has been cleaned up and documented

PR #841, opened 2026-03-29
Hotfix Daphne startup and Python runtime

Theoabw:


Dockerfile
```diff
@@ -1,5 +1,12 @@
-FROM python:3.14-alpine
-RUN apk add --no-cache gcc musl-dev libffi-dev libldap libsasl libressl bash 
+FROM python:3.12-slim
```
jaspnas, line 1: I don't think going to debian is the correct choice here, at least not to an older python version

requirements.txt
```diff
@@ -23,7 +23,10 @@ django-tables2
 django-filter
 django-bootstrap3
 django_cleanup
-daphne
+daphne==4.2.1
+autobahn==25.12.2
+Twisted==25.5.0
+zope.interface==8.0.1
```
jaspnas, line 29: These will most likely be forgotten

PR #850, opened 2026-04-02
Use non-prefixed URLs as default, language via cookie and Accept-Language

jaspnas:
## Summary

- `prefix_default_language=False`: default language (sv) now routes without a URL prefix — `/events/` instead of `/sv/events/`
- Language state is determined by cookie → `Accept-Language` header → default (sv) for non-prefixed requests (handled by Django's `LocaleMiddleware`)
- Language switcher sets the cookie and redirects back to the same non-prefixed URL instead of navigating to `/en/` or `/fi/` paths
- `/en/` and `/fi/` prefixed routes still exist and work if accessed directly

## SEO note

No language-prefixed routes for the default language. Without hreflang tags or a sitemap the benefit of keeping `/sv/` URLs is not significant enough to justify the complexity. The `<html lang="...">` attribute combined with `Accept-Language` handling is the primary language signal.

## Test plan

- [x] All 110 existing tests pass (`date-test`)
- [ ] Browse `/` and `/events/` — renders in Swedish by default
- [ ] Set `Accept-Language: en` (e.g. via curl or browser settings) on a non-prefixed URL — renders in English
- [ ] Switch language via the picker — stays at same non-prefixed URL, cookie is updated, page re-renders in selected language
- [ ] Browse `/en/events/` directly — still works (English via URL prefix)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ date-start-detached
 Important environment flags:
 
 - `PROJECT_NAME` selects the active association/site variant (`date`, `kk`, `biocum`, `demo`, `on`, ...).
-- `ENABLE_LANGUAGE_FEATURES=True` enables the localized URL prefixes, language switcher, and translated admin tabs. When omitted or false, the project runs Swedish-only.
+- `ENABLE_LANGUAGE_FEATURES=True` enables the language switcher, translated admin tabs, and runtime selection between Swedish, English, and Finnish on unprefixed URLs. When omitted or false, the project runs Swedish-only.
 - `USE_S3` toggles whether uploads use local disk storage or the configured S3-compatible backend.
 
 The script exports `COMPOSE_FILE_PATH` and defines the `date-*` aliases used throughout this README:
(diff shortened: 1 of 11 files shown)


Theoabw (dismissed):
This change does not yet implement cookie/Accept-Language-driven unprefixed routing.

1. Setting `prefix_default_language=False` only removes the prefix for the default language. With Django `LocaleMiddleware`, every unprefixed request is still resolved as `settings.LANGUAGE_CODE`, so redirecting from `set_language()` to `/...` causes users who selected `en` or `fi` to land on Swedish content.

2. The updated tests now assert redirects to `/`, but they do not verify that `/` or other unprefixed routes actually render in the selected non-default language. That means the current test suite passes while the intended behavior is still broken.

I verified this locally on the PR branch: an unprefixed request to `/` resolves to `sv` even when both the language cookie and `Accept-Language` are set to `en` or `fi`.

Please either keep prefixed URLs for non-default languages, or change the routing/middleware approach so unprefixed URLs can genuinely resolve from cookie/header before updating the redirect logic and tests.

Theoabw (commented):
The middleware change fixes the earlier problem where unprefixed URLs always resolved to Swedish.

One remaining regression: removing `i18n_patterns` means all existing `/en/...` and `/fi/...` URLs now return 404 instead of continuing to work or redirecting to the new canonical unprefixed path.

I verified this on the current PR head with URL resolution:
- `/news/` resolves normally
- `/en/news/` -> 404
- `/fi/news/` -> 404
- `/en/` -> 404
- `/fi/` -> 404

That breaks old bookmarks, shared links, and indexed URLs. Please add compatibility handling for legacy prefixed routes, ideally a redirect to the equivalent unprefixed path while preserving the selected language via cookie.

Theoabw:
Closing this.

The goal is already achieved on develop: / resolves language via cookie/header and redirects to /sv/..., /en/..., or /fi/....

This PR changes the model by removing prefixed URLs as the canonical structure. I don’t think that’s an improvement. Prefixed routes give us explicit, stable, SEO-friendly URLs, and we should keep them as the primary structure rather than relying on implicit state.

Since the current setup already covers the use case without these tradeoffs, I don’t see a reason to proceed with this change.

PR #861, opened 2026-04-03
feat: add optional 2fa for members

Theoabw:


theoabw:
I totally didn't push changes to the wrong branch before this

jaspnas:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

jaspnas:
### Dead code in `members/two_factor.py`

Two unused items introduced in this PR:

**1. `StrictTOTPDeviceForm` is never used in the setup flow**

https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/blob/695708ac0ebcbb16af8e3fae1fe8ab11ae97bc3a/members/two_factor.py#L28-L33

`MemberSetupView.done()` finds the TOTP form via `isinstance(form, TOTPDeviceForm)`, which matches the library's default `TOTPDeviceForm` — not `StrictTOTPDeviceForm`. The custom subclass is never instantiated by the view, so the `tolerance=1` setting has no effect in production. The only test for it (`test_strict_totp_form_saves_device_with_one_step_tolerance`) instantiates the form directly without going through the view, so it passes regardless.

If the intent is to enforce `tolerance=1` on enrolled devices, `MemberSetupView` needs to override `get_form_list()` to substitute `StrictTOTPDeviceForm` for `TOTPDeviceForm` in the wizard's form list. Otherwise, the class and its test should be removed.

**2. `from django import forms` is unused**

https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/blob/695708ac0ebcbb16af8e3fae1fe8ab11ae97bc3a/members/two_factor.py#L3-L3

`forms` is imported but not referenced anywhere in the file.

🤖 Generated with [Claude Code](https://claude.ai/code)

jaspnas:
### Duplicate URL registrations

All 6 URLconfs include both `members.urls` and `members.auth_urls` at the same `members/` prefix:

https://github.com/Datateknologerna-vid-Abo-Akademi/date-website/blob/695708ac0ebcbb16af8e3fae1fe8ab11ae97bc3a/core/urls/date.py#L31-L32

`members/auth_urls.py` registers the same 8 routes (`login/`, `logout/`, `password_reset/`, `password_change/`, etc.) that already exist in `members/urls.py`. Django silently uses the first match, so the second include is dead code. Any future change to one file that isn't mirrored in the other will cause subtle divergence.

`members/auth_urls.py` can be deleted entirely. The auth routes are already present in `members/urls.py` under the `members` namespace. Each URLconf just needs the duplicate line removed:

```python
# Before
path('members/', include('members.urls')),
path('members/', include('members.auth_urls')),  # duplicate — remove

# After
path('members/', include('members.urls')),
```

This applies to `core/urls/date.py`, `biocum.py`, `kk.py`, `on.py`, `demo.py`, and `pulterit.py`.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #864, opened 2026-04-04
feat: OAuth2/OIDC provider and granular permission matrix

jaspnas:
## Summary

- Turns date-website into an OAuth2/OIDC provider (RS256 JWT access tokens) via `django-oauth-toolkit`
- Adds a data-driven permission matrix: `MembershipType → Feature`, configurable in admin, embedded as a scope-filtered `permissions` claim in every JWT (e.g. `scope=upload` → only `archive.*` permissions, preventing bloat with large feature sets)
- RSA signing keys are auto-generated and stored in the DB on first startup — no manual env var management. All Docker replicas share keys via the shared DB. Key rotation via `date-manage rotate_signing_key` + container restart.
- Browser/session auth is unchanged; JWT Bearer tokens are for DRF API endpoints only
- Mounts `/o/` endpoints: token, authorize, JWKS (`.well-known/jwks.json`), OIDC discovery, userinfo, revoke

## ⚠️ Migrations need to be regenerated

Migrations `0012` and `0013` in `members/` were written by hand because Docker was not available during development. Before merging, regenerate them properly:

```bash
# Delete the manual migrations
rm members/migrations/0012_feature_signingkey_membershiptypepermission.py
rm members/migrations/0013_seed_features.py

# Regenerate
date-makemigrations members
date-manage makemigrations members --empty --name seed_features
# Then re-add the seed_features RunPython operation to the empty migration
```

## ⚠️ JWT access token claims need verification

`get_additional_claims` in `members/oauth.py` populates **ID tokens**. To also embed `permissions` in **JWT access tokens** (needed by the Go function), inspect the installed DOT source for the right override method in `JWTAccessTokenGenerator` and subclass it.

## Test plan

- [ ] Regenerate migrations and run `date-migrate`
- [ ] Start site, verify admin → Members → Signeringsnycklar shows 1 auto-generated key
- [ ] `GET /o/.well-known/openid-configuration/` returns JSON discovery doc
- [ ] `GET /o/.well-known/jwks.json` returns RSA public key with `kid`
- [ ] Create an OAuth2 Application in admin (client_credentials, scope `upload`)
- [ ] `POST /o/token/` → JWT; decode at jwt.io and verify `permissions` and `membership_type` claims
- [ ] `date-manage rotate_signing_key` + restart → JWKS shows 2 keys
- [ ] Existing session-based login and admin access still work

### DIFF
--- a/core/urls/date.py
+++ b/core/urls/date.py
@@ -38,6 +38,7 @@ urlpatterns = build_urlpatterns(
     path('polls/', include('polls.urls')),
     path('ctf/', include('ctf.urls')),
     path('admin/', admin_site.urls),
+    path('o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
     path("ckeditor5/", include('django_ckeditor_5.urls')),
     path('publications/', include('publications.urls')),
     path('alumni/', include('alumni.urls')),
--- a/archive/views.py
+++ b/archive/views.py
@@ -153,7 +153,6 @@ class FilteredExamsListView(UserPassesTestMixin, SingleTableMixin, FilterView):
 def picture_detail(request, year, album):
     collection = Collection.objects.filter(type="Pictures", pub_date__year=year, title=album).order_by(
         '-pub_date').first()
-    # TODO: get a member object and check user.is_authenticated
     if collection.hide_for_gulis and not request.user.has_feature_permission('archive.view_restricted'):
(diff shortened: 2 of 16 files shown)


core/urls/date.py
jaspnas, line 1: Need to update the other associations as well if we need it for them

PR #885, opened 2026-04-07
Upgrade S3 to s3v4, make CDN and ACL provider-agnostic

Theoabw:
- Explicit s3v4 signature version and path-style addressing to fix SignatureDoesNotMatch errors on DigitalOcean Spaces presigned URLs
- Add S3_USE_ACL toggle (default True) so ACLs can be disabled for providers like Backblaze B2 that don't support per-object ACLs
- Add endpoint_url to STORAGES OPTIONS for explicitness
- Replace hardcoded DO CDN_URL_TRANSFORMATIONS with S3_CDN_ORIGIN / S3_CDN_DOMAIN env vars, making provider switching config-only

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -93,11 +93,23 @@ S3_SECRET_KEY="password"
 # S3 bucket name
 S3_BUCKET_NAME="date-images"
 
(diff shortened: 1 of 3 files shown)


theoabw:
#888 will take care of this most likely

Issue #813, opened 2025-12-02
fix: make env vars load through docker compose instead of needing to source them


Theoabw:
**Is your fix request related to a problem?**

Environment variables are currently sourced manually before starting the project, which makes them easy to mix up or forget. This has already led to misconfigurations in production.

**Describe the solution you'd like**

Load environment variables via an env_file in Docker Compose so they are applied consistently without manual sourcing.

Theoabw:
done in #923

PR #923, opened 2026-04-21
Simplify Compose environment loading

Theoabw:
## Summary

- simplify Docker Compose environment loading so `.env` is the project source of truth
- move dev/prod Compose selection into `COMPOSE_FILE` in the example env files
- reduce duplicated environment mappings in the Compose files and remove prod host port exposure
- add a production-ready `.env.prod.example`
- update helper scripts, backup/update scripts, CI, and docs to match the new env flow
- add a shell alias installer for developers who want persistent `date-*` helpers

## Why

The previous workflow depended on sourcing `env.sh` to load environment variables into the shell. That made switching between dev and prod fragile when old exported values were still present. The new flow lets Docker Compose read `.env` directly and keeps `env.sh` focused on optional shell aliases.

## Validation

- `docker compose config --quiet` with the development example env
- `docker compose -f docker-compose.dev-all.yml config --quiet` with the development example env
- `docker compose config --quiet` with the production example env
- `python manage.py check`
- `bash -n` for the updated shell scripts
- `git diff --check`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,13 +11,13 @@
 # Sets docker compose project name to use
 COMPOSE_PROJECT_NAME="datewebsite"
 
+# Compose file used by plain `docker compose ...` commands
+COMPOSE_FILE="docker-compose.yml"
+
 # [Postgres]
 
 # PostgreSQL version to use. DO NOT change this manually after the database has been created!!!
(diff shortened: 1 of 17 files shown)


Theoabw:
Closes #813

PR #925, opened 2026-04-22
feat: add optional unfold admin theme

Theoabw:


Theoabw:
Improved admin styles and UX

image

PR #939, opened 2026-04-30
Clean up event signup flow

theoabw:
Recreated against main after develop was retired. Original PR: #936

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+
+if command -v uv >/dev/null 2>&1; then
+    echo "Running tests with uv..."
+    uv run python manage.py test
+elif docker info >/dev/null 2>&1; then
+    echo "Running tests in Docker (uv not found locally)..."
+    docker compose run --rm --no-deps web uv run python manage.py test
+else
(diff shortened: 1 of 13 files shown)


theoabw:
A few housekeeping changes landed on this branch alongside the signup flow work:

**Pre-push test hook** (`.githooks/pre-push`)
Runs the full test suite before every `git push`. Tries `uv` first (fast, local), falls back to Docker if uv is not available. Run `date-setup` once after cloning to activate it.

**`date-setup` alias** (`env.sh`)
New alias that handles the full initial setup in one command: copies `.env.example` to `.env` if it does not exist, installs the git hooks, then runs `clean_init.sh --yes` (non-interactive) to build images, run migrations, and load fixtures.

`clean_init.sh` gains a `-y`/`--yes` flag to support this: it skips both interactive prompts and defaults to not deleting media files.

**Quieter test output** (`core/settings/test.py`)
Test settings now set the root log level to `CRITICAL`, so the intentional INFO/WARNING noise from tests that exercise broken storage, failed invoices, etc. no longer clutters the test runner output. The existing `test_django_console_handler_does_not_override_logger_level` test is updated to check `core.settings.common.LOGGING` directly rather than the active settings, since it is a production config invariant.

PR #946, opened 2026-04-30
feat: kansli booking

theoabw:
Recreated against main after develop was retired. Original PR: #859


theoabw:
Sorry @brilliany, I'll probably take over this PR and complete it since this would benefit us (Datörs) quite a bit

PR #948, opened 2026-04-30
refactor: standardize UI components and modularize CSS

Theoabw:
Recreated against main after develop was retired. Original PR: #825


theoabw:
Will retry this at some point

PR #982, opened 2026-05-11
Added Pulterit functionary name entry, role translations, and homepage partner link

MaxTheMan-1216:
Made Functionary.member nullable — officials can now be added by name without a registered member account
Added name CharField to Functionary model with a clean() validation requiring either a member or a name
New migration: 0013_functionary_name_alter_functionary_member
Updated admin to show both fields with autocomplete
Fixed FunctionaryForm._post_clean() so model validation runs correctly

Changed role filter dropdown to use pk instead of title as the option value — fixes a bug where filtering by role would silently fail in non-Swedish languages after adding modeltranslation
Updated get_selected_role in functionary.py accordingly
Fixed get_filtered_functionaries to use year=year for single years and year__in=year for querysets instead of the incorrect year__in=[year]

Added members/translation.py to register FunctionaryRole.title with django-modeltranslation, generating title_sv, title_en, title_fi columns

Added a "Vill du samarbeta med Pulterit?" link pointing to /pages/yritysyhteistyo/ under the sponsors list
Styled with --accentColor (#BB29BB)
Fixed more-news link color in homepage.css to use --accentColor

Updated all three locale files (sv, en, fi) with new and corrected entries
Changed English terminology: "Functionaries" → "Officials" throughout


Theoabw (approved):
lgtm

PR #992, opened 2026-05-14
Fixed Pulterit officials page filter visibility and event sign-up field requirement

MaxTheMan-1216:
- Removed the {% if user.is_authenticated %} gate on the officials page filter form so year/role filtering is visible to all visitors, not just logged-in users.
- Fixed EventCreationForm in events/forms.py so that sign-up time fields (sign_up_others, sign_up_members, sign_up_deadline, sign_up_cancelling_deadline) are no longer required when creating an event with registration disabled.


theoabw:
lgtm, but just a small note for the future: Naming the PRs association specific when the change affects all associations can lead to a bit of confusion on the intention of the change. This change will help all association tho so it's all good

Theoabw (approved):
lgtm

PR #996, opened 2026-05-18
Improve admin UX for publications and related apps

Theoabw:


theoabw:
requires manual visual check

PR #1024, opened 2026-07-19
Fix password reset email NoReverseMatch

jaspnas:
## Summary
- Fix `NoReverseMatch: Reverse for 'password_reset_confirm' not found` when submitting the password reset form
- Django 6 removed the deprecated fallback that resolved un-namespaced URL names across all namespaces — the email template needed the `members:` namespace prefix
- Add e2e test that exercises the full password reset POST flow

## Root cause
Commit adb545b (Upgrade to django 6) upgraded from Django 5.2 to 6.0 but didn't update the password reset email template to use the namespaced URL.

## Test plan
- [ ] Submit password reset form with valid email, verify no 500 error
- [ ] Verify the reset link in the email points to the correct URL

### DIFF
--- a/members/tests.py
+++ b/members/tests.py
@@ -21,6 +21,29 @@ from members.two_factor import (
 )
 
 
+class PasswordResetFlowTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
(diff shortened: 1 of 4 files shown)


theoabw (approved):
lgtm

PR #1034, opened 2026-08-02
fix: fixed pulterit officials requiring users to be members in order to filter officials

MaxTheMan-1216:
- Fixed except ValueError, TypeError: in functionaries/selectors.py that was a hard SyntaxError under Python 3, breaking the entire public functionaries (/members/funktionarer/) page.
- Removed the request.user.is_authenticated requirement on the year/role filters — the listing itself is already fully public, and not every hosted association has member accounts, so gating the filter behind login served no purpose and silently broke filtering for logged-out visitors.
- Updated the two tests that had locked in the old "ignore filters for anonymous users" behavior to assert the new, correct behavior instead.

### DIFF
--- a/functionaries/selectors.py
+++ b/functionaries/selectors.py
@@ -23,7 +23,7 @@ def get_selected_year(request, distinct_years):
     selected_year = get_current_year()
     all_years = False
 
-    if request.user.is_authenticated and 'year' in request.GET:
+    if 'year' in request.GET:
         year_param = request.GET['year']
 
         if year_param == 'all':
@@ -44,7 +44,7 @@ def get_selected_role(request, functionary_roles):
     selected_role = None
     all_roles = False
 
-    if request.user.is_authenticated and 'role' in request.GET:
+    if 'role' in request.GET:
         role_param = request.GET['role']
 
         if role_param == 'all':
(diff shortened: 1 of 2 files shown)


functionaries/selectors.py
```diff
@@ -23,7 +23,7 @@ def get_selected_year(request, distinct_years):
     selected_year = get_current_year()
     all_years = False
 
-    if request.user.is_authenticated and 'year' in request.GET:
+    if 'year' in request.GET:
```
jaspnas, line 26: I think this was intentionally set to require authentication to view previous year's functionaries. Will possibly impact DaTe

functionaries/selectors.py
```diff
@@ -53,7 +53,9 @@ def get_selected_role(request, functionary_roles):
         else:
             try:
                 role_id = int(role_param)
-            except ValueError, TypeError:
+            except ValueError:
```
jaspnas, line 56: Why not have this as tuple to reduce code duplication?

functionaries/selectors.py
jaspnas: If desired for Pulterit we might want to put this behind a feature flag and keeping the old behaviour as default. Also worth thinking about GDPR

jaspnas (changes requested):

functionaries/selectors.py
jaspnas: If it's due to formatter we might want to configure it

theoabw:
Also please try to use conventional titles in the PRs

PR #1041, opened 2026-08-05
fix: Support multiple events per day in homepage calendar (#489)

### LINKED ISSUE #489: bug:Only one event per day can be opened from calendar
**Describe the bug**  
Calendar opens events by day specific ID, which is not unique

**To Reproduce**  
Create two events for a single day and try opening them from calendar

**Additional context**  
Calendar only shows the latter event and defaults to it when clicking on the calendar day

theoabw:
## Summary

Fixes the homepage calendar only exposing one event per day (#489).

The calendar payload keyed every `YYYY-MM-DD` to a **single** event, so when two events shared a day the later one overwrote the earlier, and clicking the day always opened the last event.

## Changes

- `date/views.py::format_calendar_events` now maps each date to a **list** of events (link, title, start time).
- Calendar JS (all 6 association variants: date/sf, kk, biocum, demo, pulterit, impuls) builds one popup containing a link per event for days with several events. Days with a single event keep the existing direct-navigation on click.
- Removed the duplicated `calendar_link` id from popup links (multiple links per day would repeat the id; styling uses the `.calendar-eventday-popup` class only).
- Fixed a latent bug in the biocum/demo variants where hours were never zero-padded.

## Testing

- `date-test`: 80 tests pass, incl. a new test asserting both events on the same day are kept in order.
- `date-manage check`: clean.
- `djlint --check` on all changed templates: clean.

Closes #489

### DIFF
--- a/templates/date/date/components/scripts.html
+++ b/templates/date/date/components/scripts.html
@@ -25,43 +25,46 @@
         }
     };
 
-    Object.entries(events_json).forEach((o)=>{
-        const tempDate = new Date(o[1].eventFullDate);
-        let hours = tempDate.getHours();
-        let minutes = tempDate.getMinutes();
-        // Ensure hours and minutes are displayed with two digits
-        hours = hours < 10 ? '0' + hours : hours;
-        minutes = minutes < 10 ? '0' + minutes : minutes;
-        const rawUrl = (o[1].link || '').trim();
-        
-        const link = document.createElement('a');
-        link.className = "calendar-eventday-popup";
-        link.id = "calendar_link";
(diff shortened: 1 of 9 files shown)


templates/date/date/components/scripts.html
jaspnas, line 1: Should all these scripts files be merged into one?

theoabw: Good call. The calendar script was identical in all six, so it is one shared partial now. I kept the files separate since the tails differ (joke box, partner carousel, nothing).

PR #1043, opened 2026-08-05
fix: Update JS dependencies and add integrity checks (#689)

### LINKED ISSUE #689: fix: Update javascript dependencies and add integrity checks

Js dependencies are old, and they don't have integrity checks

Go through what needs to be updated and add integrity checks to all js imports

theoabw:
## Summary

Closes #689: old JS dependencies were vendored/loaded without integrity checks.

## Changes

**Integrity checks (SRI)** — every remaining CDN import now carries `integrity` + `crossorigin`:
- masonry-layout + imagesloaded (archive detail)
- line-awesome CSS (publications list/index/password/viewer, sf base)
- pdf.js viewer CSS (publications viewer)
- Already covered before: bootstrap, bootstrap-icons, jquery (base), glightbox, reconnecting-websocket

**Dependency updates:**
- `events/list.html`: jQuery **1.7.1** (Google CDN, 2011) → jQuery 3.7.1 with SRI
- KK/biocum homepages: vendored jQuery **1.11.3** (2015) → vendored jQuery 3.7.1 (full build, for jCarouselLite's animate calls)
- `base.html` + 8 event templates: jQuery slim 3.4.1 → 3.7.1 (one version site-wide)
- Vendored jCarouselLite patched for jQuery 3: `.size()` → `.length` (removed in jQuery 3.0)
- sf base: unpinned icons8 line-awesome URL → pinned cdnjs 1.3.0 build with SRI

**Follow-up done in this PR:** vendored vanilla-calendar updated from the old v2.x build to **v2.9.10** (last v2 release): plain CSS, unchanged visuals, `clickDay` now reads the clicked date from `date.selectedDates[0]` (v2.9 API change) in all six association variants.

**Still out of scope:** v3 (`vanilla-calendar-pro`) is ESM-only and ships its styles as Tailwind `@apply` source, requiring a build pipeline and a calendar integration rework — a separate migration. Turnstile `api.js` and dynamic pdf.js module imports can't use SRI by browser design.

## Testing

- `date-test`: full suite passes (also confirms kk/biocum/sf templates render)
- `date-manage check`: clean
- `djlint --check templates/`: 0 files would be updated
- Patched JS validated with `node --check`; SRI hashes computed from the exact CDN bytes (jQuery 3.7.1 hashes match the officially published values)

Closes #689


theoabw:
needs visual confirmation

jaspnas (approved):
LGTM if it works out of the box. I recall there being some issues with jquery version previously.

theoabw:
Rebased onto `main` and verified against the merged #1041.

- The six `*/date/components/scripts.html` conflicts are resolved by keeping `main`'s version. Those files are now thin wrappers around `templates/common/date/partials/calendar_scripts.html`, and that shared partial already reads `date.selectedDates[0]`, so this branch's own edits to those six files are now redundant.
- `templates/sf/core/base.html` pointed at the plain Line Awesome build, which only defines `la/las/lar`. Shared templates use `fa-*`, `fas` and `far`, so sf rendered no icons. It now uses the same `font-awesome-line-awesome` stylesheet and hash as the shared base template. Added a test that renders `core/base.html` for every association, since the existing check only ran under the default variant and never exercised sf's override.
- On the review note about the new stylesheet following the OS dark preference: v2's `themeDetect` targets `html[data-theme]`, which this project never sets, so `system` falls through to `light`. Verified in a DOM with the real bundle: no dark rules can activate, and `v2`'s `data-calendar-theme` resolves to `light`.

Also verified the library swap works with the shared handler using the real bundle in jsdom: `VanillaCalendar` global is exported, `init()` builds the calendar, popups get `modifier`/`html`, a single-event day navigates, and a multi-event day opens the popup.

One cosmetic follow-up not addressed here: v2's stylesheet sets `width:300px` on `.vanilla-calendar`, where the previous hand-written theme used `width:100%`, so the homepage calendar no longer stretches to its column. Say the word if you want that pinned in this PR.

PR #1148, opened 2026-08-30
fix: admin media static paths; parallelize build-time collectstatic

theoabw:
## Problem

Admin pages that use translated-field tabs (staticpages nav, events, ...)
and the publications collection-access widget 500 when logged in:

```
ValueError: The file 'common/css/admin_translation_tabs.css' could not be found
```

The admin `Media` references pointed at `common/...` paths, but the
per-variant static layout (`static/<variant>` + `static/common` in
`STATICFILES_DIRS`) collects `static/common` contents without the prefix;
every template in the codebase uses the unprefixed form (e.g.
`events/css/style.css`). The errors were masked until the unfold static
fix (#1147) by the unfold 500, which fired first on every admin render.

## Changes

- `core/admin.py`, `publications/admin.py`: drop the `common/` prefix from
  the three Media references; update the two tests asserting the old path.
- `Dockerfile`: run the per-variant collectstatic loop in parallel and drop
  the dev-only `demo` variant (no collected tree needed in the prod image).

## Verification

- Local amd64 build: collectstatic step 69.7s (sequential, 7 variants) ->
  15.3s (parallel, 6 variants).
- Manifests: `css/admin_translation_tabs.css`,
  `publications/js/admin-collection-access.js` present on all variants;
  unfold keys only on date/pulterit/sf/impuls (36 each), none on kk/biocum.
- Runtime resolution of all previously broken paths returns hashed URLs.
- `core.tests.test_admin_ux`, `core.tests.test_modeltranslation`,
  `publications` tests: 36 tests pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,21 +26,25 @@ RUN python manage.py compilemessages -l en -l fi -l sv
 # collection. Settings pick the tree matching the runtime PROJECT_NAME.
 ARG PROJECT_NAME=date
 ENV PROJECT_NAME=$PROJECT_NAME
-# Prod runs the unfold admin skin (USE_UNFOLD=True via extraEnv) on the date
-# (incl. qa), pulterit, sf and impuls variants; kk, biocum and demo keep the
-# stock admin. collectstatic only discovers app static dirs for apps in
-# INSTALLED_APPS, so unfold's files must be collected with USE_UNFOLD=True or
-# every unfold admin template render 500s on the missing manifest entry
-# (2026-08-30).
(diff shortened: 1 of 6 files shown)


theoabw:
Addressed review findings:

- Added a regression test (core/tests/test_modeltranslation.py) that pins the layout contract: admin Media paths must be unprefixed (the STATICFILES_DIRS layout strips the common/ prefix) and the source file must exist under static/common. Covers LanguageTabbedTranslationAdmin and PDFFileAdmin. The old assertions passed even with the broken paths.
- Qualified the kubernetes.md statement: only production associations get collected trees; the dev-only demo variant gets none.

PR #1150, opened 2026-08-30
fix: member autocomplete for editors without members.view_member

jaspnas:
## Problem

You can't manually set **Solver** on a flag: the member dropdown reports *"no results"* no matter what you type.

The `solver` field (and the `FlagInline`) uses Django's admin autocomplete. `AutocompleteJsonView` only answers the AJAX request when the requesting user holds **view permission on the related model** (`members.Member`). CTF editors are staff via a group that carries the `ctf.*` permissions but *not* `members.view_member` (members hold PII — phone, address — and the registry is deliberately restricted), so the endpoint returns `403` and Select2 renders an empty dropdown.

Same issue anywhere an editor group manages objects that point at `Member` (or `EventAttendees`) through `autocomplete_fields` without also holding the related model's view permission:

| Admin | Field |
| --- | --- |
| `FlagAdmin` / `FlagInline` | `solver` |
| `GuessAdmin` | `user` |
| `EventAdmin` | `author` |
| `EventAttendeesAdmin` | `avec_for` |
| `EventInvoiceAdmin` | `participant` |
| `PostAdmin` (news) | `author` |
| `FunctionaryAdmin` / `FunctionaryInline` | `member` |

## Fix

`core.admin.ReferringObjectAutocompleteJsonView`, installed via `FixedLanguageAdminSite.autocomplete_view`, widens the permission check: in addition to Django's "view permission on the related model", it accepts the request when the user may **add or change the object that declares the autocomplete field** (checking the registered `ModelAdmin` and any matching inline). If you can edit a Flag, you can search members to pick its solver.

The related `ModelAdmin` still owns `get_queryset` / `get_search_results`, so `UserAdmin`'s `MEMBER_ADMIN_RESTRICTED_GROUP` / `MEMBER_ADMIN_RESTRICTED_MEMBERSHIP_TYPE` row filtering keeps applying to the returned choices.

## Tests

New `core/tests/test_admin_autocomplete.py` — a group with only `ctf.change_flag` (resp. `functionaries.change_functionary`, `events.change_event`, `billing.change_eventinvoice`) gets member/attendee results; a group with only an unrelated permission still gets `403`; `members.view_member` and superuser paths unchanged.

- `date-test core.tests ctf events.tests functionaries billing members.tests` — 338 pass
- `ruff check .`, `ruff format --check .`, `mypy .`, `manage.py check` — clean

## Docs

Updated `docs/dev/members.md`, `docs/dev/ctf.md`, `docs/admin/ctf.md`.


theoabw (approved):
lgtm

PR #1152, opened 2026-08-31
Revert "k8s: serve WebSockets from the web deployment, drop the separate asgi pod (#1146)"

jaspnas:
## Summary

Reverts #1146. HTTP serving goes back to a **WSGI web tier** (`gunicorn core.wsgi`) with a **separate Daphne ASGI process** for WebSockets. Chart `0.4.0 -> 0.5.0`.

#1146 has been causing intermittent `500` responses in QA since it merged, including on pages that exist. This rolls back to the last known-good topology so the site is stable; the ASGI consolidation can be retried on a branch with load testing.

## Symptom

Intermittent `500`s with this traceback tail, on both missing URLs and valid ones (`/events/<slug>/`, `/media/...`, `/robots.txt`):

```
psycopg2.InterfaceError: connection already closed
django.db.utils.InterfaceError: connection already closed
```

The 404 path is hit hardest because a failed 404 render cascades into a failed 500 render, but valid event pages 500 too.

## Root cause

#1146 switched the web deployment from `core.wsgi` to `core.asgi:application` (gunicorn + `uvicorn_worker.UvicornWorker`), so one process pool serves HTTP and `/ws/`.

Under Django's ASGI handler, every HTTP request wraps its synchronous work in a per-request `ThreadSensitiveContext` (`asgiref/sync.py`). That context creates a **single-thread executor per request and shuts it down when the request finishes** — middleware, the view, template rendering and the `handler404` / `handler500` error views all run on that short-lived thread.

DB connections are thread-local. With `CONN_MAX_AGE = 600` (`core/settings/common.py`), Django keeps the connection open across requests and only closes it once it is "obsolete". At the end of an ASGI request the connection is ~1s old, so `close_old_connections` leaves it open — and then the executor thread is destroyed with a live psycopg2 connection that Django was never told to close.

Consequences under load:

- PostgreSQL backends leak (one per request that outlives its thread), drifting toward `max_connections`.
- PostgreSQL / the pooler starts terminating idle connections server-side.
- `CONN_HEALTH_CHECKS = True` does not save the request: the health check runs off the `request_started` signal, whose sync receivers run on a different thread than the exception path (`response_for_exception` is dispatched `thread_sensitive=False`), so a poisoned connection on that path is used without a health check → `InterfaceError: connection already closed` → `500`.

WSGI does not have this problem: sync work runs on gunicorn's own worker thread for the whole request/response cycle, inside Django's normal connection lifecycle.

### Why it started "today"

`4d1d6f8` (#1146) is the only change between the previous QA image and the one showing the errors. Before it, `web` ran `core.wsgi` and a separate `asgi` pod ran Daphne for `/ws/` only.

## What this PR does

Straight `git revert` of `4d1d6f8`, i.e. the tree returns to the state of #1150 (`85b7013`), plus:

- **Chart `0.4.0 -> 0.5.0`** (not back to `0.3.12`): the revert also reverts the operator-visible values interface — the `asgi:` block and the `/ws` ingress/gateway route come back — so it is a minor bump in the same class as `0.4.0`, and the published chart version never moves backwards. No major bump: the chart is pre-1.0 and this is not an API-stability commitment.
- Restores `media.s3.staticEnabled` / `staticCustomDomain` in `values.schema.json`, which #1146 removed even though `deployment-web.yaml` still references `.Values.media.s3.staticEnabled`.

### Operator action required

Per-site values that dropped their `asgi:` overrides after `0.4.0` need them back, and the `/ws` route returns to the ingress/gateway. (These are the same values that existed on `0.3.x`.)

### Also reverted — re-land separately

#1146 bundled unrelated improvements that this revert also undoes:

- `core/settings/test.py` — keep Django loggers audible at `CRITICAL` instead of dropping them
- `core/tests/test_asgi.py` — ASGI HTTP smoke test
- `uvicorn` / `uvicorn-worker` dependency + the `uvicorn.workers` → `uvicorn_worker` migration

## Follow-up

Retrying the single-pool ASGI topology needs, at minimum: `CONN_MAX_AGE = 0` (or a real connection pool) for the ASGI path, `close_old_connections()` in `handler404` / `handler500`, and a k6 run that watches PostgreSQL connection count and 5xx rate through sustained traffic.

## Testing

- `uv run ruff check .` / `ruff format --check .` — pass
- `uv lock --check` — consistent
- `DJANGO_SETTINGS_MODULE=core.settings.test manage.py check` — clean
- `manage.py test core date.tests` — 241 pass (3 skipped)


theoabw:
Superseded by #1153 (merged 2026-08-31, deployed to all 7 sites): the ASGI topology is kept and the 500s are fixed at the source (executor-thread CloseConnectionsMiddleware + CONN_MAX_AGE=0 + error-handler connection closes). Verified live: 0 connection errors after rollout. Closing without merging.

PR #1153, opened 2026-08-31
fix: close DB connections per request under the ASGI handler (executor-thread middleware)

theoabw:
## Summary

Fixes the intermittent 500s introduced by the asgi merge (#1146) without reverting the single-pool topology.

## Root cause (two layers)

1. Under Django's ASGI handler, sync middleware and views run on asgiref executor threads (thread-local DB connections), while the `request_finished` signal is sent **asynchronously on the event loop thread** (`signals.request_finished.asend` in `handlers/asgi.py`). `close_old_connections` therefore never runs on the thread that owns the connection.
2. Every request leaks its PostgreSQL connection on the shared executor thread. When the server reaps the idle connection (`idle_session_timeout`), the next request on that thread 500s with `psycopg2.InterfaceError: connection already closed` / `server closed the connection unexpectedly`. `CONN_MAX_AGE` and `CONN_HEALTH_CHECKS` cannot fix this: no cleanup ever runs on the owning thread.

Verified in production (2026-08-31): with `CONN_MAX_AGE=0` set cluster-wide, kk still 500ed; `pg_stat_activity` showed request connections left idle on the web pod for minutes, and error log timestamps correlated with ~10-minute-old reaped connections.

## Changes

- `date/middleware.py`: new `CloseConnectionsMiddleware` — calls `close_old_connections()` in a `finally` after `get_response`, running on the same executor thread as the views. Registered outermost after WhiteNoise in `core/settings/common.py`.
- `core/settings/common.py`: `CONN_MAX_AGE` default 600 -> 0 (`DB_CONN_MAX_AGE` env override stays) as the same policy for paths outside the middleware and for long-lived processes, which can opt back in.
- `date/views.py`: `handler404` / `handler500` call `close_old_connections()` before rendering, so the error page never reuses a poisoned connection mid-render.
- `docker-compose.prod.yml`: web service pins `DB_CONN_MAX_AGE: "0"` (wins over `.env`).
- `.env.example`, `docs/dev/kubernetes.md` updated.
- Tests: middleware closes connections after success and after a raised view; error handlers close before rendering; settings guard pins the 0 default.

## Operator note

Per-site operator values currently set `extraEnv: DB_CONN_MAX_AGE: '60'`; those go to `'0'` (done, values-only, already rolled 2026-08-31). Note the values-only change does NOT fix the 500s: the middleware is the actual fix and ships with this image (auto-update rolls it to all sites).

## Testing

- `ruff check` / `ruff format --check` clean
- `manage.py check` clean
- `manage.py test date core`: 259 pass, 3 skipped

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,8 +23,10 @@ DATE_POSTGRESQL_VERSION=16
 DATE_DB_PASSWORD="password"
 
 # How long (seconds) a database connection is kept alive between requests.
-# 0 disables persistent connections.
-DB_CONN_MAX_AGE=600
+# 0 (default) closes it at request end: the web tier serves ASGI, where each
+# request runs on a per-request thread and persistent connections leak.
+# Long-lived processes (celery) may set a positive value.
+DB_CONN_MAX_AGE=0
 
 # [pgAdmin]
 
--- a/core/settings/common.py
+++ b/core/settings/common.py
@@ -285,10 +285,17 @@ DATABASES = {
         'PASSWORD': env_alias('DATE_DB_PASSWORD', 'DB_PASSWORD', default=''),
         'HOST': env('DB_HOST', str, 'db'),
         'PORT': env('DB_PORT', int, 5432),
-        # Keep PostgreSQL connections alive between requests to avoid
-        # connection setup latency; must live inside the database config.
-        'CONN_MAX_AGE': env('DB_CONN_MAX_AGE', int, 600),
-        # Re-verify persistent connections so they survive database restarts.
+        # Close the connection at the end of every request (0). The web tier
+        # serves ASGI (core.asgi), where each request's sync work runs on a
+        # per-request executor thread that asgiref destroys when the request
+        # finishes; a positive CONN_MAX_AGE leaves the thread-local connection
+        # open on a dead thread, leaking PostgreSQL backends and 500ing later
+        # requests with "connection already closed" (2026-08-31). Long-lived
+        # processes (celery workers, management commands) can opt back into
+        # persistent connections via DB_CONN_MAX_AGE.
+        'CONN_MAX_AGE': env('DB_CONN_MAX_AGE', int, 0),
+        # Re-verify persistent connections so they survive database restarts
+        # (only applies when DB_CONN_MAX_AGE is positive).
         'CONN_HEALTH_CHECKS': True,
     }
 }
--- a/core/settings/common.py
+++ b/core/settings/common.py
@@ -228,6 +228,10 @@ COMMON_CONTEXT_PROCESSORS = [
 
 MIDDLEWARE = [
     'whitenoise.middleware.WhiteNoiseMiddleware',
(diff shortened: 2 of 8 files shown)


jaspnas:
Reviewed the mechanism against the Django 6.0 handlers in the lockfile. The approach is sound and I'm satisfied it targets the actual root cause:

- **`CONN_MAX_AGE=0` fixes the normal path.** `ASGIHandler.handle()` runs entirely inside `async with ThreadSensitiveContext()`, and `response.close()` is dispatched with `sync_to_async(response.close)()` (thread-sensitive), so it runs on the *same* per-request executor thread as the view. Its `request_finished` -> `close_old_connections()` fires before asgiref tears that thread down, and with `CONN_MAX_AGE=0` the connection is always obsolete so it actually closes. No connection is left on a dead thread.
- **`close_old_connections()` in the error handlers plugs the exception path.** `response_for_exception` is dispatched `thread_sensitive=False` (`django/core/handlers/exception.py`), i.e. on a pooled thread with no `request_started`/`request_finished` bracketing, so a stale connection there is never health-checked. Closing at the top of `handler404`/`handler500` is the right fix.

Smaller blast radius than a full revert, and it keeps the single-process topology. A few things before merge:

**1. Make the chart self-sufficient.** `charts/` never templates `DB_CONN_MAX_AGE`, so on Kubernetes the new default only takes effect if nothing in the deployment env overrides it. `docker-compose.prod.yml` pins `web` explicitly; the chart should do the same - template `DB_CONN_MAX_AGE: "0"` into the web configmap/deployment (with a values key) so the fix ships with the chart and isn't dependent on external env. Otherwise merging this doesn't change cluster behaviour until a separate deploy-config change lands.

**2. celery / long-lived processes.** The compose `celery` service and the chart celery/beat deployments inherit the `0` default and lose persistent connections + the `CONN_HEALTH_CHECKS` behaviour they were designed around. Either set `DB_CONN_MAX_AGE` (e.g. 600) explicitly for those, or note in the PR that celery now uses per-task connections and that's intended.

**3. Load evidence.** The regression is load- and timing-dependent ("intermittent", "sometimes even when the event exists"). Current testing is unit tests + `manage.py check`, so the fix is confirmed by reasoning only. A k6/soak run (the `perf/load-tests` scripts) showing `pg_stat_activity` flat and 5xx at zero under sustained load against both a 404 path (`/media/...`) and a valid `/events/<slug>/` would actually demonstrate it.

**4. Residual, minor.** The `thread_sensitive=False` exception-pool threads still never get `request_finished`, so a connection opened in `handler500` lingers on that pooled thread until the next error's `close_old_connections()`. Bounded by pool size and self-healing - fine to leave, but worth a one-line note in the handler comment.

**5. Nit.** `handler500` has no comment while `handler404` does - hoist the comment above both or cross-reference.

**6.** The "Operator note" in the description quotes specifics from the private deploy repo; suggest trimming it to something like "the deployment env override for `DB_CONN_MAX_AGE` must be cleared for this to take effect" and keeping the details in the infra-repo change.

PR #1169, opened 2026-09-04
fix: auto-reconnect the live attendee WebSocket

theoabw:
## Summary

The event-page live attendee list opens one WebSocket and, until now, gave up permanently when it closed: `onclose` only logged an error, so the list silently went stale until a manual reload.

Sockets now close routinely: gunicorn worker recycling (being enabled in the operator deployment to bound the web pods' slow memory drift) and deploys terminate the worker holding the socket. This makes those drops self-healing.

## Changes

- reconnect with capped exponential backoff (1s doubling to 30s), reset on a successful open
- no retry stacking: one socket at a time, and no retry after the page unloads (`beforeunload` closes the socket and flags the page as leaving)
- `onerror` closes the socket so the close handler owns the retry path

## Validation

- `node --check static/common/core/js/live-attendee-list.js`
- logic unchanged on the message path (same handler body); UI behavior identical on a stable connection

### DIFF
--- a/docs/dev/events.md
+++ b/docs/dev/events.md
@@ -46,9 +46,12 @@
 - Websocket notifications: `ws_send(slug, form, public_info)` broadcasts new
   registrations (except during tests). The slug is derived from the parent event
   if present. The client (`static/common/core/js/live-attendee-list.js`)
-  auto-reconnects with capped exponential backoff (1s doubling to 30s) after a
-  close or error, so worker recycling, deploys and network blips do not leave
-  the live list stale.
+  auto-reconnects with capped exponential backoff (1s doubling to 30s, reset
+  only after a connection has stayed open) after a close or error, so worker
+  recycling, deploys and network blips do not leave the live list stale.
+  Broadcasts received while disconnected are not replayed and the row counter
+  is local, so the list can show gaps after a longer outage until the page is
+  reloaded.
 - Billing hook: when `settings.EXPERIMENTAL_FEATURES` contains `event_billing`,
   `billing.handlers.handle_event_billing()` runs after a successful signup to
   generate invoices or send confirmations.
--- a/docs/dev/events.md
+++ b/docs/dev/events.md
@@ -45,7 +45,10 @@
   Turnstile responses in `cf-turnstile-response`.
 - Websocket notifications: `ws_send(slug, form, public_info)` broadcasts new
   registrations (except during tests). The slug is derived from the parent event
-  if present.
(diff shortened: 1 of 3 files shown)


theoabw:
Follow-up review fixes:
- parentSlug branch now connects to the only existing route (`/ws/events/<slug>/`); child-event pages previously built `/ws/<slug>/`, which never matched the URL router, so their live list could not connect at all (pre-existing, now visible because reconnect would otherwise retry a dead endpoint forever).
- pageshow (back/forward cache restore) resets the closedByPage flag and reconnects if the socket died during the freeze.
- Verified against the server side: the consumer always accepts (no reject-then-close path), so the 1s backoff reset on open cannot turn into a hot loop; permanent failures stay bounded at the 30s cap.

theoabw:
Second review round (general agent) findings, fixed in 14b901f:

1. Double-connect race (blocker): a pending retry timer from onclose could race a pageshow-triggered connect(), and connect() unconditionally created a socket even when one was OPEN/CONNECTING -> two live sockets, each broadcast appended twice. Fixed: connect() is now single-flight (clears any pending retry timer, returns when a socket is already OPEN/CONNECTING).
2. Backoff reset on every accept: a worker that accepts and then dies kept the client at the 1s retry floor for the whole churn. The backoff now resets only after a connection has stayed open 5s.
3. Idle sockets on pages without list markup: the script now exits early when neither #attendee-list nor #attendees exists (kk100 index pages, sign_up-off detail pages); tomtejakt pages get a working live list for the first time.
4. Accepted and documented: broadcasts received while disconnected are not replayed and the row counter is local, so gaps/misnumbering can occur after longer outages until reload (noted in docs/dev/events.md and a code comment). No partial-list endpoint exists to resync from; out of scope for this PR.

theoabw:
Third review round (review agent) findings, fixed in commit 4:

1. tomtejakt template lacked the #attendee-list container / data-parent-slug every other live-list template carries: a child event rendered through it would subscribe to its own slug while ws_send broadcasts to the parent group. The container is now normalized (same id + data-parent-slug), so all eight live-list templates feed the JS the same way.
2. A pending reconnect timer survived navigation: beforeunload now cancels it, and connect() refuses to run while closedByPage is set (belt and braces for a delayed timer firing during teardown).
3. The markup gate was loose (either marker allowed execution): pages rendering #attendee-list without a table (signup off, list hidden for old events) kept an idle, reconnecting socket. The gate is now the #attendees table itself.

Validated: node --check, djlint on the changed template.

PR #1170, opened 2026-09-05
style: styling tweaks to the Impuls site

MaxTheMan-1216:
Made CSS adjustments, logo placements, and added credits section similarly to the pulterit site.


theoabw (commented):
The extras block now exposes two competing reporting channels. The "Trakasseriombud" heading link (extras.html:5) still points at this site's internal `/social/harassment/` form, while the paragraph below sends reports to the external URL. If all Impuls reports should go through the form page, both links should point there; otherwise make clear that the internal ombud form is intentionally kept.

Also: impulsrf.org is this site's own domain, and `/pages/trakasseriformular/` is a route on this same app (staticpages slug page, `pages` is included in the impuls URLconf). The hardcoded absolute URL is a self-link: it makes dev/staging instances point at production and bypasses language handling. Use a same-site reference, e.g. `{{ '/pages/trakasseriformular/'|localized_url }}` or a URL reverse, and drop `target="_blank"` for a same-site link.

templates/impuls/date/components/extras.html
```diff
@@ -6,7 +6,9 @@ <h5 class="heading-link">
   </h5>
   <p>
     {% trans "Om du upplever trakasserier i samband med Impuls verksamhet, vänligen" %}
-    <a href="{% url 'social:harassment' %}">{% trans "rapportera" %}</a>
+    <a href="https://impulsrf.org/pages/trakasseriformular/"
```
theoabw: The "Trakasseriombud" heading link just above (line 5) still points at this site internal `/social/harassment/` form, so this block now offers two competing report channels. If all Impuls reports should go through the form page, update that link too.

Also, impulsrf.org is this site own domain and `/pages/trakasseriformular/` is a route on this same app (staticpages slug page), so the hardcoded absolute URL is a self-link: dev/staging instances would point at production and language handling is bypassed. Use a same-site reference such as `{{ /pages/trakasseriformular/|localized_url }}` or a URL reverse, and drop `target="_blank"` for a same-site link.

static/impuls/date/css/homepage.css
```diff
@@ -563,13 +567,17 @@ div.calendar-container a {
   box-shadow: 0 0 0 0;
 }
 
-.vanilla-calendar-day__btn_today {
+/* "Today" only gets its own highlight when it has no event that day —
+ * matching the site's own background makes it read as an absence of the
+ * usual white day-circle rather than a competing color. When an event
+ * also lands on today, .calendar-eventday below takes over instead. */
+.vanilla-calendar-day__btn_today:not(.calendar-eventday) {
```
theoabw, line 574: A day that is both today and has an event loses its today marker: calendar-eventday is applied to the same day button, so this :not() drops the today styling and the event styling takes over entirely. Keep a distinct affordance (border, outline, or inset shadow) for the combined state.

static/impuls/core/css/base.css
```diff
@@ -26,6 +54,17 @@ html body{
     border-radius: var(--defaultRadius);
 }
 
+/* Only images the editor wrapped in a link (CKEditor: image -> link
+ * properties) get the hover effect, same as the footer partner logo. */
+.content a img {
+    transition: transform 0.2s, opacity 0.2s;
+}
+
+.content a:hover img {
```
theoabw, line 61: This applies to every editor image wrapped in a link, not just partner-style logos. A full-width image already at its max can grow 3 percent past its content box, risking clipping or horizontal overflow, and the opacity dip may be unexpected on ordinary linked illustrations. Consider an opt-in class or constraining overflow.

static/impuls/date/css/homepage.css
```diff
@@ -317,11 +317,15 @@ div.news-content a:hover{
 }
 
 .more-news{
-    color: var(--linkColorDark);
+    /* Sits directly on the navy page background, not a white card, so this
+     * uses --linkColorDeepHover rather than --linkColorDeep: the fully dark
+     * shade only clears ~1.6:1 contrast here, well under WCAG's 3:1 minimum
+     * for text. --linkColorDeepHover already clears 3.8:1. */
+    color: var(--linkColorDeepHover);
```
theoabw, line 324: A var named for the hover state now serves as the resting color, while the actual hover uses the hardcoded #7f95b4 just below. Contrast is fine, but a dedicated variable for the lighter shade would keep palette maintenance sane.

static/impuls/core/css/base.css
```diff
@@ -13,10 +13,38 @@ html body{
 }
 
 .association-footer-brand .footer-img {
-    max-height: 150px;
+    max-height: 160px;
     object-fit: contain;
 }
 
+.footer-partner-logo svg {
+    max-height: 155px;   /* was 155px — this is the one knob to eyeball-tune */
```
theoabw: "was 155px", "eyeball-tune", and "see earlier fix" read as review-time commentary, not durable documentation; the first is also self-contradictory since the declared value is still 155px.

templates/impuls/core/footer.html
```diff
@@ -1,7 +1,34 @@
...
+    <small>{% trans "Website developed and maintained by" %}  <a class="footer-credit-link"
+     href="https://datateknologerna.org/"
+     target="_blank"
+     rel="noopener">DaTe</a>
+      <br>
+      <a class="footer-credit-link"
+         href="https://datateknologerna.org/"
+         target="_blank"
+         rel="noopener">
+        <img src="{% static 'core/images/Sp_Albin.png' %}"
```
theoabw, line 29: This PNG is byte-identical to the pulterit copy (and the one added for sf in #1171). Since every association's static dirs fall back to static/common/, a single copy there would serve all three sites.

theoabw (approved):
lgtm

PR #1171, opened 2026-09-05
feat: sf date credits

MaxTheMan-1216:
Added the same credits section as for the impuls/pulterit sites in accordance with the signed website contract.


theoabw (commented):
Nit: `static/sf/core/images/Sp_Albin.png` is 148 KB at 756x742 px but renders at max 40px tall; a properly sized asset would cut needless bytes. It is also byte-identical to the pulterit copy (and the one added for impuls in #1170); a single copy under `static/common/core/images/` could serve all three.

Otherwise verified: markup matches the pulterit credits block, translations exist in all catalogs, static path and CSS vars resolve for SF, links are `noopener` with alt text. No blockers.

theoabw (approved):
lgtm

PR #1180, opened 2026-09-18
Add public JSON endpoint for upcoming events

jaspnas:
## Summary
- Adds a public, unauthenticated JSON endpoint at `GET /api/events/upcoming/` (`api:events:upcoming`) listing published, upcoming events with their public details and times.
- Introduces `API_ROUTES` in `core/urls/common.py`: a shared global `/api/` root, assembled in `build_urlpatterns()` only from route keys that were both requested **and** have an `API_ROUTES` entry. Disabling a capability for an association (by leaving its key out, e.g. no `'events'`) removes its API surface too — no separate toggle to remember, and no import of that app's `api_urls` module when the capability isn't enabled.
- Excludes `members_only` events, matching `EventDetailView`'s existing behavior of blocking anonymous access to those detail pages.
- Each entry includes `id`, `title`, `slug`, absolute `url`, ISO-8601 `event_date_start`/`event_date_end`, `content`, and `image`. Responses carry a 60s public `Cache-Control` header.
- Unreleased endpoint, so no backwards-compat alias for an older path was needed.

## Test plan
- [x] `date-test` — full suite passes (one pre-existing, unrelated local-env failure in `test_database_settings` confirmed present on `main` too)
- [x] New `ApiRootCompositionTests` proving the `/api/` root appears/disappears with the `'events'` route key
- [x] Updated `core/tests/test_url_route_parity.py` expectations for all 7 associations
- [x] `date-manage check` across all 7 association settings modules (date, kk, biocum, demo, pulterit, sf, impuls)
- [x] `uv run ruff check .` / `uv run ruff format --check .`
- [x] `uv run mypy .`
- [x] Manual smoke test: started the dev stack and curled `GET /api/events/upcoming/`, got a 200 with the expected JSON


events/api.py
```diff
@@ -0,0 +1,42 @@
...
+    """Public JSON list of upcoming, published, non-members-only events.
+
+    Members-only events are excluded: anonymous visitors cannot view their
+    detail page (see EventDetailView._requires_member_login), so they are
+    left out of this public listing too.
+    """
+    now = timezone.now()
+    events = (
+        Event.objects.published()
+        .filter(event_date_end__gte=now, members_only=False)
```
theoabw: Severity: high

`members_only` is filtered here but the passcode gate is not, so a passcode-protected event is published in full. `EventDetailView` swaps in `events/event_passcode.html` whenever `event.passcode` is set and the session has not validated it (events/views.py:137-157): the HTML page hides the content, while this endpoint returns it. I reproduced that with a probe, 200 with the passcode form and no content on the detail page, and the same event's `content` returned verbatim here. `exclude(passcode="")` plus a test would close it. If including passcode events is deliberate, please say so in the docstring and in docs/dev/events.md, since both currently imply this listing matches the detail view's gates. For context, /events/feed already exposes published events with tags stripped, members-only ones included, so the pattern predates this PR.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+        "url": request.build_absolute_uri(reverse('events:detail', args=[event.slug])),
+        "event_date_start": event.event_date_start.isoformat(),
+        "event_date_end": event.event_date_end.isoformat(),
+        "content": event.content,
+        "image": event.background_image_url,
+    }
+
+
+@require_GET
+@cache_control(public=True, max_age=UPCOMING_EVENTS_CACHE_SECONDS)
```
theoabw, line 38: Severity: medium

`Cache-Control: public, max-age=60` is only safe if the body is identical for every client, and this body is not. `title` and `content` are modeltranslation fields and the language is read from the `django_language` cookie, because `USE_ACCEPT_LANGUAGE_HEADER` is False for every variant (date/middleware.py:53-64). In a probe, two requests to the same URL with different language cookies returned different bodies while both carried `Cache-Control: public, max-age=60` and `Vary: Accept-Language` with no `Vary: Cookie`, so a shared cache in front of the site can hand a Swedish body to an English visitor. Adding `Vary: Cookie` fixes the correctness, and dropping `public=True` is the simpler alternative.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+
+def _serialize_event(request, event):
+    return {
+        "id": event.pk,
+        "title": event.title,
+        "slug": event.slug,
+        "url": request.build_absolute_uri(reverse('events:detail', args=[event.slug])),
+        "event_date_start": event.event_date_start.isoformat(),
+        "event_date_end": event.event_date_end.isoformat(),
+        "content": event.content,
```
theoabw, line 24: Severity: medium

Language is cookie-driven on this deployment and `Accept-Language` is ignored, so a machine client has no way to ask for English here and gets Swedish unless it forges the cookie. A `?lang=` parameter validated against `settings.LANGUAGES`, or honouring `Accept-Language` for API requests, would make this usable for the consumers an API is for.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+        "image": event.background_image_url,
```
theoabw: Severity: medium

`url` is absolute but `image` is not. `background_image_url` returns the storage URL as-is, which is `/media/...` under local storage (a probe returned `"/media/events/test.png"`) and absolute only with S3. `request.build_absolute_uri(event.background_image_url)` is a no-op on absolute values and makes the two fields consistent.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+
+UPCOMING_EVENTS_CACHE_SECONDS = 60
+
+
+def _serialize_event(request, event):
+    return {
+        "id": event.pk,
+        "title": event.title,
+        "slug": event.slug,
+        "url": request.build_absolute_uri(reverse('events:detail', args=[event.slug])),
```
theoabw, line 21: Severity: low

When an event has `redirect_link` set, this `url` is a detail page that immediately 302s off-site (EventDetailView.get) and `content` tends to be empty, so a consumer following `url` never arrives on your site. Exposing the redirect target as its own field, or documenting this, would help.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+        "event_date_start": event.event_date_start.isoformat(),
+        "event_date_end": event.event_date_end.isoformat(),
+        "content": event.content,
+        "image": event.background_image_url,
+    }
+
+
+@require_GET
```
theoabw: Severity: low

`require_GET` answers 405 to HEAD (I checked), and some clients and uptime monitors send HEAD first. `require_safe` allows GET and HEAD while still rejecting writes.

events/api.py
```diff
@@ -0,0 +1,42 @@
...
+    detail page (see EventDetailView._requires_member_login), so they are
+    left out of this public listing too.
+    """
+    now = timezone.now()
+    events = (
+        Event.objects.published()
+        .filter(event_date_end__gte=now, members_only=False)
+        .exclude(slug="")
+        .exclude(slug__isnull=True)
+        .order_by('event_date_start')
```
theoabw: Severity: low

`order_by('event_date_start')` is not deterministic when events share a start time, so `('event_date_start', 'id')` would make the order stable for consumers that diff responses. There is also no cap on the list, which is fine at the current volume, but a limit would keep a public endpoint from growing without bound.

events/tests.py
```diff
@@ -156,6 +156,55 @@ def test_get_event_detail(self):
...
+        entry = payload["events"][0]
+        self.assertEqual(entry["slug"], self.event.slug)
+        self.assertEqual(entry["title"], self.event.title)
+        self.assertEqual(
+            entry["event_date_start"],
+            self.event.event_date_start.isoformat(),
+        )
+        self.assertTrue(entry["url"].endswith(reverse('events:detail', args=[self.event.slug])))
+
+    def test_upcoming_events_api_does_not_require_authentication(self):
```
theoabw: Severity: low

This test only asserts the status. In `setUp` the event's `event_date_end` is effectively now, so by the time the request runs the filter usually returns an empty list and the test passes whether or not the endpoint lists anything. Asserting the payload would make it a real authentication test. Two behaviours are also untested: the passcode exclusion above and the `Cache-Control` header. A dedicated class for the API tests would also keep them out of the signup test case.

core/urls/common.py
```diff
@@ -32,6 +32,17 @@
...
 
+# JSON API routes, keyed by the same route name as the capability they belong
+# to. A key here is only ever included when that same key was also requested
+# from ROUTES, so disabling a capability for a variant (by leaving its key
+# out of build_urlpatterns' arguments) removes its API surface for free and
+# never imports that app's api_urls module. Collected under one shared
+# `/api/` root (namespace `api`) rather than as one-off per-app routes, so
+# `reverse()` targets look like `api:events:upcoming`.
+API_ROUTES = {
+    'events': lambda: path('events/', include('events.api_urls')),
```
theoabw, line 51: Severity: low, and it is a question rather than a defect.

Worth confirming this is the intended shape: keying `API_ROUTES` to the same keys as `ROUTES` means a variant cannot serve its HTML event pages without also publishing this JSON API, and there is no API-only key. The laziness and the single-toggle property are good; I mainly want the case of an association that wants the pages but not the API to be a decision rather than a side effect.

docs/dev/events.md
```diff
@@ -147,6 +147,23 @@
...
   overrides can be triggered by event title or slug for special layouts.
+- `events/api.py::upcoming_events` (`GET /api/events/upcoming/`, route name
+  `api:events:upcoming`) is a public, unauthenticated JSON endpoint listing
+  published events with `event_date_end` in the future. It is wired in from
+  `events/api_urls.py` via `core.urls.common.API_ROUTES['events']`. The
+  shared `/api/` root is assembled automatically in `build_urlpatterns()`
+  from whichever `API_ROUTES` keys were also requested from `ROUTES`, so an
+  association that leaves `'events'` out of its route list gets neither the
+  HTML routes nor `events.api_urls` imported — no separate toggle to
+  remember. See that dict's docstring in `core/urls/common.py` before adding
```
theoabw: Severity: nit

`API_ROUTES` carries a `#` comment rather than a docstring, so pointing readers at "that dict's docstring" will not resolve. A clause noting that `content` is raw CKEditor HTML would also help, since the .ics feed strips tags and consumers may otherwise assume plain text.

theoabw (commented):
Reviewed the branch head and ran the events suite (142 tests), the URL parity test (3), ruff, djlint and mypy locally, plus probe tests for the passcode, cache, language and image behaviour commented below.

High: the passcode gate, which I would close before merge. Medium: the cache header, language handling and the image URL shape. Low or nit: everything else.

This review was done by an AI agent.

events/api.py
jaspnas: Fixed in 70ee148: added `passcode=""` to the filter alongside `members_only=False`, and added a test (`test_excludes_passcode_protected_event`).

events/api.py
jaspnas: Fixed in 70ee148: added `@vary_on_cookie`, verified live (`Vary: Cookie, Accept-Language`) and via `test_sets_public_cache_control_with_vary_cookie`.

events/api.py
jaspnas: Fixed in 70ee148: added an optional `?lang=` query param, validated against `settings.LANGUAGES` and applied via `translation.override(...)`. Covered by `test_lang_parameter_overrides_cookie_language` / `test_unknown_lang_parameter_is_ignored`.

events/api.py
jaspnas: Fixed in 70ee148: `image` now goes through `request.build_absolute_uri(...)` like `url`, guarded against calling it on an empty string.

events/api.py
jaspnas: Fixed in 70ee148: added `redirect_link` as its own field rather than only documenting the behavior.

events/api.py
jaspnas: Fixed in 70ee148: switched `require_GET` to `require_safe`, verified HEAD now returns 200 live.

events/api.py
jaspnas: Fixed in 70ee148: ordering is now `(event_date_start, id)`, and results are capped at 200 (`MAX_RESULTS`).

events/tests.py
jaspnas: Fixed in 70ee148: moved the API tests into their own `EventsUpcomingApiTests` with a real future-dated event fixture, and the auth test now asserts the actual payload slug instead of just status 200. Also added passcode-exclusion and Cache-Control/Vary coverage per the other comments.

core/urls/common.py
jaspnas: Good question, and deliberate: today no association wants an app's HTML pages without its API. Added a comment above `API_ROUTES` in core/urls/common.py spelling that out, and noting the escape hatch (a second, distinct key pair, the way 'archive' vs 'archive_exams' already split one app) if that need ever comes up, rather than adding an unused flag now.

docs/dev/events.md
jaspnas: Fixed in 70ee148: corrected the docs wording (it's a comment, not a docstring) and added a note that `content` is raw CKEditor HTML, unlike the .ics feed which strips tags.

events/api.py
```diff
@@ -0,0 +1,64 @@
...
+    `?lang=` query parameter validated against `settings.LANGUAGES`, for
+    clients that cannot carry a cookie.
+    """
+    now = timezone.now()
+    events = (
+        Event.objects.published()
+        .filter(event_date_end__gte=now, members_only=False, passcode="")
+        .exclude(slug="")
+        .exclude(slug__isnull=True)
+        .order_by('event_date_start', 'id')[:MAX_RESULTS]
```
theoabw: Severity: nit

The 200 cap is silent, so a client cannot tell a complete list from a truncated one. A `count` or `truncated` field in the payload would make that visible; the cap itself is documented and the `id` tiebreak keeps the boundary stable.

events/api.py
```diff
@@ -0,0 +1,64 @@
...
+def _serialize_event(request, event):
+    return {
+        "id": event.pk,
+        "title": event.title,
+        "slug": event.slug,
+        "url": request.build_absolute_uri(reverse('events:detail', args=[event.slug])),
+        "event_date_start": event.event_date_start.isoformat(),
+        "event_date_end": event.event_date_end.isoformat(),
+        "content": event.content,
+        "image": request.build_absolute_uri(event.background_image_url) if event.background_image_url else "",
```
theoabw: Severity: nit

`event.background_image_url` is evaluated twice, so an image whose storage URL raises gets its warning logged twice. A local variable avoids the repeated lookup.

events/tests.py
```diff
@@ -545,6 +545,94 @@ def test_billing_runs_after_signup_commit(self, mock_handle_event_billing):
...
+    def test_lang_parameter_overrides_cookie_language(self):
+        self.event.title_en = 'English title'
+        self.event.title_sv = 'Svensk titel'
+        self.event.save()
+
+        response = self.client.get(reverse('api:events:upcoming'), {'lang': 'en'})
+
+        self.assertEqual(response.json()["events"][0]["title"], 'English title')
+
+    def test_unknown_lang_parameter_is_ignored(self):
```
theoabw, line 630: Severity: nit

This test asserts only the status, and the fallback title is deterministic, so asserting it would make the test meaningful. HEAD through `require_safe`, the 200 cap and the ordering are also covered by hand only; all three are cheap to add in this class.

theoabw (approved):
I re-ran the branch at 70ee148 and verified all ten earlier findings against the code rather than against the replies: the passcode and members-only exclusions, `Vary: Cookie`, `?lang=`, the absolute `image`, `redirect_link`, HEAD through `require_safe`, the ordering tiebreak and the 200 cap, the new `EventsUpcomingApiTests`, the `API_ROUTES` rationale comment, and the docs. Full suite 669 tests, ruff, ruff format, djlint and mypy pass locally, and CI is green on this commit.

The three nits inline do not block this. Two notes that sit outside the diff: on deployments with `ENABLE_LANGUAGE_FEATURES=False`, `settings.LANGUAGES` holds only `sv`, so `?lang=` is ignored there by design; and `events/feed.py` still publishes members-only and passcode-protected events with tags stripped, which deserves its own issue if passcode gating is meant to be enforced.

This review was done by an AI agent.

events/api.py
jaspnas: Fixed in fc0107f: added a `truncated` boolean to the top-level response (fetch MAX_RESULTS+1, trim, and report whether the extra row existed), documented in docs/dev/events.md. Covered by `test_response_is_not_truncated_below_the_cap` / `test_response_reports_truncation_above_the_cap`.

events/api.py
jaspnas: Fixed in fc0107f: `_serialize_event` now reads `event.background_image_url` into a local once.

events/tests.py
jaspnas: Fixed in fc0107f: `test_unknown_lang_parameter_is_ignored` now asserts the fallback title, and added `test_head_request_is_allowed`, `test_orders_tied_events_by_id`, and the two truncation tests above.

PR #1185, opened 2026-09-22
Drop stale edition prefix from kk Baal page heading

jaspnas:
## Summary
- The kk Baal event detail page (`templates/common/events/baal_detail.html`) had its `<h1>` hardcoded to "CII Kemistbaal" (a stale Roman-numeral edition prefix). Changed it to just "Kemistbaal".
- Scoped to the standard yearly Baal template only; the separate 100th-anniversary template (`kk100_detail.html`, "100 Kemistbaal") is untouched.

## Test plan
- [x] `uv run djlint --check templates/common/events/baal_detail.html`
- [x] Manual smoke test: rendered an event through this template locally and confirmed the heading shows "Kemistbaal"


theoabw (approved):
lgtm if kk wants this. kk100 could be removed tbh, but idk if there's some specific reason to keep it

jaspnas:
> lgtm if kk wants this. kk100 could be removed tbh, but idk if there's some specific reason to keep it

Don't think so, can be recovered from git anyway. But let's scope it to a separate PR

